PR #1, opened 2025-11-24
Buildstream skeleton

alatiera:


alatiera:
We also need this merged https://gitlab.gnome.org/GNOME/gnome-build-meta/-/merge_requests/4311

alatiera:
oh also, we need to add the rest of the shell extensions. I only did two of them as an example.

PR #19, opened 2026-02-09
Generate updatable tarball

### LINKED ISSUE #12: Switch from using ublue-os/packages for homebrew
the `ublue-brew` rpm package in [ublue-os/packages](https://github.com/ublue-os/packages) hasn't been updated in 2 months, and the tarball that is still being uploaded weekly should probably not be used anymore.

jumpyvi:
Fix #12 

This generate a tarball from the official brew releases (compatible with the ublue brew install script).

The git stuff is so it can be updated with "brew update"

I need help with some testing before this can be merged

### DIFF
--- a/elements/bluefin/brew-tarball.bst
+++ b/elements/bluefin/brew-tarball.bst
@@ -1,22 +1,37 @@
 kind: manual
 
 build-depends:
-  - freedesktop-sdk.bst:public-stacks/runtime-minimal.bst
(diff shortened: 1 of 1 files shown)


alatiera (changes requested):

hecknt:
In its current state, when you launch a terminal for the first time, it'll do a download for portable-ruby.

image

I think more setup has to be done on brew before it gets thrown into a tarball.

PR #20, opened 2026-02-14
Overwrite nautilus for Ptyxis support

jumpyvi:
Need to be tested first, works in my sysext builds

### DIFF
--- a/elements/core/meta-gnome-core-apps.bst
+++ b/elements/core/meta-gnome-core-apps.bst
@@ -8,6 +8,6 @@ depends:
   - gnome-build-meta.bst:core/baobab.bst
   - gnome-build-meta.bst:core/gnome-disk-utility.bst
   - gnome-build-meta.bst:core/gnome-system-monitor.bst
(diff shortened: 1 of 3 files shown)


elements/core/meta-gnome-core-apps.bst
```diff
@@ -8,6 +8,6 @@ depends:
   - gnome-build-meta.bst:core/baobab.bst
   - gnome-build-meta.bst:core/gnome-disk-utility.bst
   - gnome-build-meta.bst:core/gnome-system-monitor.bst
-  - gnome-build-meta.bst:core/nautilus.bst
   - gnome-build-meta.bst:sdk/yelp.bst
+  - core/nautilus.bst
```
alatiera: you also need to add this in the override section in `gnome-build-meta.bst`

jumpyvi: Oh yeah thx.

Anyways I think Ghostty is the default term now?

And I would like to make it work with xdg-open-terminal instead

PR #37, opened 2026-02-19
feat: add zsh as default login shell and config

hanthor:
This PR adds zsh to the image and sets it as the default login shell for new users.


castrojo:
I'm +1 on this but is a significant enough change that we should get feedback from the others on this one!

castrojo:
Concensus in discord was that it wasn't worth adding the work needed to support this in projectbluefin/common - especially since we explicitly cleaned things up. Maybe someday.

PR #66, opened 2026-02-27
feat(sysext): WIP sysext/confext build target

ahmedadan:
## Summary

- Adds a parallel build target (`sysext/layer.bst`) that produces Bluefin as a systemd system extension (sysext) + configuration extension (confext) layered on top of vanilla GNOME OS
- Adds `build-sysext` and `export-sysext` Justfile recipes
- Adds sysupdate `.transfer` and `.feature` files for update delivery
- Existing OCI pipeline is completely untouched

## What works

- `just build-sysext` builds successfully (all 1066 elements)
- `just export-sysext` produces two DDI images:
  - `bluefin-x86-64_l.1.raw` (419MB sysext — `/usr` delta)
  - `bluefin-confext-x86-64_l.1.raw` (4MB confext — `/etc` delta)

## Known limitations / TODOs

- [ ] Signing keys use snakeoil `kind: local` — need to vendor the `generated` plugin for build-time key generation
- ~`image-version` and `filesystem-time` are hardcoded in `layer.bst` — should be inherited from junction or set by CI~
- [x] `image-version` and `filesystem-time` defined in `project.conf` via `include/image-version.yml`
- [ ] `make-layer` whiteout creation uses a Python monkey-patch workaround for rootless podman (no `CAP_MKNOD` in user namespaces) — works but not elegant
- [ ] CDN URL in `.transfer` files (`dl.projectbluefin.io`) is a placeholder
- [ ] Plymouth theme included in sysext but won't take effect (initramfs is baked into host UKI)
- [ ] Needs VM testing: copy DDIs to GNOME OS, run `systemd-sysext merge` + `systemd-confext merge`
- [ ] Homebrew tarball (~500MB) inflates the sysext DDI — may need splitting later

## Test plan

- [x] `just build-sysext` completes without errors
- [x] `just export-sysext` produces `.raw` files
- [x] `just bst show oci/bluefin.bst` still resolves (no regression)
- [x] `systemd-dissect` inspection of DDI images
- [ ] VM integration test on GNOME OS

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,10 @@
 *.tar
 .ovmf-vars.fd
 .build-out
+.sysext-out
+
+# Snakeoil signing keys (generate locally, not committed)
+files/sysext-keys/
 
 # Local AI agent state (opencode sessions, drafts)
 # Skills are committed; everything else is ignored
--- a/Justfile
+++ b/Justfile
@@ -111,12 +111,37 @@ export:
     # Step: Chunkify (reorganize layers)
     just chunkify "{{image_name}}:{{image_tag}}"
 
+# ── Sysext ───────────────────────────────────────────────────────────
+# Build the Bluefin system extension (sysext + confext DDIs).
+[group('build')]
+build-sysext:
+    #!/usr/bin/env bash
+    set -euo pipefail
+
+    echo "==> Building Bluefin sysext with BuildStream..."
+    just bst build sysext/layer.bst
+    echo "==> Sysext build complete."
+
+# Export sysext artifacts from the BuildStream cache.
+# Assumes `just build-sysext` has already completed.
+[group('build')]
+export-sysext:
+    #!/usr/bin/env bash
+    set -euo pipefail
+
+    echo "==> Exporting sysext artifacts..."
+    rm -rf .sysext-out
+    just bst artifact checkout sysext/layer.bst --directory /src/.sysext-out
+
+    echo "==> Sysext artifacts:"
+    ls -lh .sysext-out/*.raw
+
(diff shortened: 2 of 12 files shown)


alatiera:
> [ ] `image-version` and `filesystem-time` are hardcoded in `layer.bst` — should be inherited from junction or set by CI

These two you can copy over in bluefin. They don't have to gnome-build-meta at all

ahmedadan:
**VM integration test results (GNOME OS nightly)**

Tested DDIs against a vanilla GNOME OS nightly install (`IMAGE_VERSION=nightly.1002649`).

**`systemd-dissect` validation — passed:**
- Both DDIs recognized as valid: `✓ sysext for system`, `✓ confext for system`
- Extension-release metadata parsed correctly (`ID=org.gnome.os`)
- Partition layout correct: erofs root + dm-verity hash + verity signature

**`systemd-sysext merge` — blocked by kernel verity policy (to the best of my understanding):**
- GNOME OS kernel has `CONFIG_DM_VERITY_VERIFY_ROOTHASH_SIG=y` — we believe this requires signing keys enrolled in the secondary/platform keyring
- Snakeoil keys can't be enrolled as they are self-signed with no chain of trust to the built-in keyring
- My attempts to bypass this via `--image-policy`, `/etc/verity.d/` (PEM and DER), and `keyctl` were all unsuccessful
- I believe merge testing requires either MOK-enrolled keys or keys signed by a trusted CA

**Takeaway:** DDI structure and metadata are correct. Merge is blocked solely on key trust, which aligns with the existing TODO to replace snakeoil keys with proper build-time key generation.

PR #101, opened 2026-03-15
bluefin: Consolidate per-arch elements

alatiera:
The only different part was the per-arch specific sources, but suing the following yaml syntax we can put the conditional inline the file.

```yaml
sources:
- kind: tar
  (?):
  - arch == "x86_64":
      url: foo
      ref: foo
  - arch == "aarch64":
      url: bar
      ref: bar
```


alatiera:
We should also redo the rest of the elements as well

castrojo (dismissed):

castrojo (approved):

PR #164, opened 2026-04-06
feat(brew): fetch Homebrew directly from upstream

dylanmtaylor:
Replace the ublue-os/packages tarball with a direct fetch from Homebrew/brew v5.1.4.

This removes the third-party dependency on ublue-os/packages weekly releases. The element clones Homebrew/brew, sets up the expected directory layout, and tars it for the existing brew-setup.service to unpack on first boot.

No longer needs per-arch elements since brew is Ruby scripts.

Ref #12


castrojo:
Can we adjust this to pull from ublue-os/brew? We need it to be consistent with the other Bluefins since it also has PATH fixes in there and stuff. We don't want to have 2 different brew experiences.

dylanmtaylor:
> Can we adjust this to pull from ublue-os/brew? We need it to be consistent with the other Bluefins since it also has PATH fixes in there and stuff. We don't want to have 2 different brew experiences.

@castrojo I think that would require https://github.com/ublue-os/brew/pull/14 be merged in, which adds a workflow to publish release tarballs that dakota can consume. Without it, there's no tarball to fetch from ublue-os/brew. The old ublue-os/packages tarballs stopped being generated in March so the current ones are frozen.

PR #192, opened 2026-04-10
feat(ci): add BuildStream build caching

dylanmtaylor:
Adds GitHub Actions caching for BuildStream artifacts between CI runs.

Two changes:

1. Fixes `cachedir` in the CI config to point to `/root/.cache/buildstream` (the path already mounted from the host via the Justfile). Previously it was `/srv/cache` which was not mounted, so the entire local cache was thrown away after every run.

2. Adds `actions/cache` to persist `~/.cache/buildstream` between runs. Cache key is based on the junction refs and deps list so it invalidates when upstream changes. Partial hits still help via restore-keys fallback.

On a warm cache this should make subsequent builds near-instant since BuildStream can skip pulling artifacts from remote servers entirely.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,6 +54,13 @@ jobs:
       - name: Setup Just
         uses: extractions/setup-just@53165ef7e734c5c07cb06b3c8e7b647c5aa16db3 # v4
 
+      - name: Cache BuildStream artifacts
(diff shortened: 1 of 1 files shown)


dylanmtaylor:
I see my local artifact cache is 42G... so this probably won't help too much.

PR #204, opened 2026-04-12
Add Bitwarden

dylanmtaylor:
Adds Bitwarden desktop app and CLI.

- bitwarden.bst: desktop app from upstream tar.gz (x86_64 + aarch64)
- bitwarden-cli.bst: CLI tool from upstream zip (x86_64 + aarch64)
- Both added to deps.bst

Same pattern as the 1Password elements.

image

### DIFF
--- a/elements/bluefin/bitwarden-cli.bst
+++ b/elements/bluefin/bitwarden-cli.bst
@@ -1,0 +1,25 @@
+kind: manual
+
+sources:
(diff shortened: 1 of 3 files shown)


dylanmtaylor:
Gemini is hallucinating based on assumptions about what the tarball contains rather than looking at the actual build results.

Issue #207, opened 2026-04-12
(Discussion)  Use brew for docker on Dakota (for now)


jumpyvi:
This is the best solution I've found to easily setup (rootfull) docker on Dakota.
I already maintain a dockerd build in the experimental tap

https://github.com/jumpyvi/ujust-extras-CE/blob/master/dxlessv3.just

`just -f dxlessv3 setup-docker-rootful` will auto-magically setup docker with brew and systemd-sysusers.

The same is possible for libvirt/qemu too, but qemu depends on brew's glibc which is annoying, we would have to make our own formulae build.

znmeb:
Testing this - I had to manually create the `docker` group as a system group and add myself to it. Then:

[docker.log.zip](https://github.com/user-attachments/files/27579807/docker.log.zip)

That's as far as I've gotten ... there may be more missing pieces. The only test case I have that uses Docker rather than Podman is the Armbian cross-build, and I don't know if it's rootless. It croaks on regular Bluefin with SELinux problems as well.

castrojo:
https://github.com/projectbluefin/common/pull/545

PR #208, opened 2026-04-12
Add Docker support (containerd + dockerd + CLI)

dylanmtaylor:
Adds Docker to the image using official static binaries (Docker 28.1.1, includes containerd + runc).

Socket-activated by default. Users need `sudo usermod -aG docker $USER` to use without sudo.


dylanmtaylor:
Closing in favor of the brew approach per #207.

PR #209, opened 2026-04-12
feat(bluefin): add uupd auto-updater element

castrojo:
Ok this is a copy of the other Bluefins except we need to build uupd of course. I can confirm that the service installs in the VM and shows the right state, etc. Though it doesn't work when built locally since that's a localhost image. 

Agent notes:

Ships the ublue-os/uupd binary (v1.3.0) as a BuildStream manual element, matching the files installed by the uupd RPM in production Bluefin:

- /usr/bin/uupd
- /usr/lib/systemd/system/uupd.service (--disable-module-distrobox, matching production Bluefin build_files/base/17-cleanup.sh)
- /usr/lib/systemd/system/uupd.timer
- /usr/lib/systemd/system/uupd-manual.service
- /etc/uupd/config.json
- /etc/polkit-1/rules.d/uupd.rules

The systemd preset (enable uupd.timer) is already provided by projectbluefin/common via common.bst — no duplicate needed.

### DIFF
--- a/elements/bluefin/deps.bst
+++ b/elements/bluefin/deps.bst
@@ -9,6 +9,7 @@ depends:
   - bluefin/brew-tarball.bst
 
   - bluefin/tailscale.bst
+  - bluefin/uupd.bst
 
   - bluefin/1password-cli.bst
   - bluefin/1password.bst
--- a/elements/bluefin/uupd.bst
+++ b/elements/bluefin/uupd.bst
@@ -1,0 +1,120 @@
+kind: manual
+
+sources:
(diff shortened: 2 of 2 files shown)


castrojo:
Some of the agent reviews might be worth investigating but I'd rather keep it as close to a copy as the other bluefins so ignoring most of them. 

I wonder if it's worth reinvestigating the uupd services in general across the board? Dunno.

castrojo:
Since the clients aren't updating anyway we might as well ship this, only so much I can do locally.

PR #210, opened 2026-04-13
feat(extensions): replace LogoMenu with Custom Command Menu

castrojo:
Ok so this menu is clean and very easy for users to extend, and we can drop the fork/rpm, so this is nice and clean. Also upstream is looking at accepting our PR to add the hostname there. So the subtle menu will have the machine hostname. Very clean. 

Swaps logomenu for StorageB/custom-command-menu, which is actively maintained, supports translations, and is fully customizable. Depends on projectbluefin/common#264.

### DIFF
--- a/elements/bluefin/shell-extensions/custom-command-menu.bst
+++ b/elements/bluefin/shell-extensions/custom-command-menu.bst
@@ -2,9 +2,9 @@ kind: manual
 
 sources:
   - kind: git_repo
-    url: github:ublue-os/Logomenu.git
-    track: main
-    ref: ab8b37e524d6795f4eb7170cc62b7f90de7f5729
+    url: github:StorageB/custom-command-menu.git
+    track: v*
+    ref: v12-0-g8749b1247e681e86e07a3c7e7db5bdcead139639
 
 build-depends:
   - freedesktop-sdk.bst:public-stacks/buildsystem-make.bst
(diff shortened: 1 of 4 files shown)


elements/bluefin/shell-extensions/custom-command-menu.bst
```diff
@@ -2,9 +2,9 @@ kind: manual
 
 sources:
   - kind: git_repo
-    url: github:ublue-os/Logomenu.git
-    track: main
-    ref: ab8b37e524d6795f4eb7170cc62b7f90de7f5729
+    url: github:StorageB/custom-command-menu.git
+    track: v*
+    ref: v12-0-g8749b1247e681e86e07a3c7e7db5bdcead139639
```
dylanmtaylor: Can we merge https://github.com/projectbluefin/common/pull/272 and use the git ref for the now merged in dynamic command feature here instead of a tag?

castrojo: Yeah leave a github change thing in the UI for this and I'll merge it.

dylanmtaylor:
```suggestion
    ref: f7d4eb671aece43c9ee8b52bf8d740f948202bf4
```

dylanmtaylor: @castrojo commented.

PR #226, opened 2026-04-14
feat: add asusctl 6.3.7 for ASUS ROG/TUF laptop support

castrojo:
This adds the components needed so when 6.19 and 7.0 kernels drop we get better Asus support. Supported by the OGC. 

The GUI we're leaving out, brew is probably the way to go for the GUI apps. 

Agent notes:

Packages the asusctl suite (asusd daemon, asusctl CLI, asus-shutdown, asusd-user) as a cargo2/make BuildStream element.

Hardware gating: upstream udev rules (99-asusd.rules) triple-gate activation via DMI vendor glob (ASUS*), product family allowlist (ROG/TUF/Zephyrus/Strix/Vivobook/Zenbook/ProArt/TX Air/Gaming/EXPERTBOOK), and asus-nb-wmi driver probe. No systemd preset installed — the service is invisible on non-ASUS hardware (never loaded, absent from list-units/--failed, zero journal noise at boot).

Build notes:
- rog-control-center excluded via Cargo.toml workspace patch (two sed patterns cover per-line members list and inline default-members array; guard fails loudly if upstream reformats)
- sg-rs (git dep of rog-scsi → asusd) vendored via cargo2 git entry
- Added pkg-config to build-depends (not in buildsystem-make.bst)
- --offline instead of --locked (workspace patch invalidates lock check)

Installs:
  /usr/bin/{asusd,asusctl,asus-shutdown,asusd-user} /usr/lib/udev/rules.d/99-asusd.rules /usr/lib/systemd/system/{asusd,asus-shutdown}.service /usr/lib/systemd/user/asusd-user.service /usr/share/dbus-1/system.d/asusd.conf /usr/share/asusd/{aura_support.ron,anime/**} /usr/share/rog-gui/layouts/*.ron /usr/share/icons/hicolor/512x512/apps/asus_notif_*.png

Verified: bst build ✅  just build ✅  just lint ✅
Binaries confirmed in final OCI image.

Future: CONFIG_ASUS_ARMOURY (battery limits, fan curves, Aura/RGB on 2021+ ROG hardware) requires kernel ≥6.19; add module to kernel config patch once freedesktop-sdk bumps.

### DIFF
--- a/elements/bluefin/asusctl.bst
+++ b/elements/bluefin/asusctl.bst
@@ -1,0 +1,2845 @@
+kind: make
+
+build-depends:
+  - freedesktop-sdk.bst:components/rust.bst
+  - freedesktop-sdk.bst:public-stacks/buildsystem-make.bst
+  - freedesktop-sdk.bst:components/pkg-config.bst
+  - freedesktop-sdk.bst:components/systemd-libs.bst
+
+depends:
+  - freedesktop-sdk.bst:public-stacks/runtime-minimal.bst
+  - freedesktop-sdk.bst:components/systemd-libs.bst
+
+config:
+  build-commands:
+    # Remove rog-control-center from workspace members so cargo does not try to resolve
+    # its git-only Slint/supergfxctl deps. Two patterns cover both forms used at 6.3.7:
+    #   members list (one entry per line) and inline default-members array.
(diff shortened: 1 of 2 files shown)


castrojo:
we got this in brew, no need!

PR #235, opened 2026-04-15
feat(deps): add bootc v1.15.1 local override with CI tracking

castrojo:
bootc in gnomeos is 1.12, we need the latest.


castrojo:
Addressed: expanded all 5 abbreviated `rev:` fields (cfsctl, composefs, composefs-boot, composefs-ioctls, composefs-oci) to full 40-char commit hashes in 48d0b55.

PR #232, opened 2026-04-15
chunkah: rechunk with user.component xattrs via extended fakecap

hanthor:
## Problem

BST artifact xattrs (`user.component`, `user.update-interval`) are stripped on OCI export. chunkah uses rustix raw syscalls for xattr reads, bypassing libc — so LD_PRELOAD shims can't intercept them. Without component xattrs chunkah falls back to bigfiles-only splitting.

See coreos/chunkah#113.

## Approach

An extended `fakecap.c` (upstreamable to freedesktop-sdk) plus a writable overlayfs:

1. `gen-filemap.py` queries the BST artifact cache → `filemap.json` + `fakecap-manifest.tsv` (path → component)
2. `chunkah-filemap.bst` builds `fakecap-restore` and bakes the manifest into the image at `/usr/lib/chunkah/`
3. `just chunkify` mounts a writable overlayfs, runs `fakecap-restore` to physically write xattrs, then runs chunkah via bind-mount

**Future path** (after coreos/chunkah#113 lands): drop the overlay entirely — run chunkah with `LD_PRELOAD=fakecap.so FAKECAP_MANIFEST=...`. `fakecap.c` already intercepts `getxattr` and serves from the manifest; only the raw-syscall bypass blocks this today.

## New files

| File | Purpose |
|---|---|
| `files/fakecap/fakecap.c` | Extended fakecap: manifest-based getxattr serving + build-time capture |
| `files/fakecap/fakecap-restore.c` | Physically writes xattrs from manifest to rootfs |
| `scripts/gen-filemap.py` | BST artifact cache → filemap.json + fakecap-manifest.tsv |
| `elements/oci/chunkah-filemap.bst` | Builds fakecap-restore, bakes manifest into image |

## Results (197k files, 7.4 GiB, 120 layers)

| Approach | Components | Unclaimed |
|---|---|---|
| Baseline | 0 | 7.3 GiB |
| This PR | 709 | 17 MiB |

## TODO
- Remove overlay + fakecap-restore step once coreos/chunkah#113 lands
- See #231 to upstream extended fakecap.c to freedesktop-sdk

PR #246, opened 2026-04-16
chunkify: enable in export and reduce max-layers to 120

hanthor:
Enables the chunkify step that was commented out in the `export` target, and sets `--max-layers 120` (down from 128) to stay safely within OCI registry layer limits.

## Changes
- Uncomment `just chunkify` in the `export` target so CI builds actually run rechunking
- Reduce `--max-layers` from 128 to 120

## Context
PR #232 added the full fakecap xattr rechunking pipeline. The `chunkify` call in `export` was left commented out pending validation. Local testing confirms it works end-to-end:
- 824 components, 120 layers, 36.2 MiB unclaimed (vs 7.3 GiB baseline with 0 components)

The max-layers reduction gives an 8-layer safety margin against OCI registry limits while keeping the same fine-grained splitting behaviour.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,8 +19,8 @@
 
 # Git worktrees
 .worktrees/
-files/filemap.json
-files/fakecap-manifest.tsv
+# files/filemap.json and files/fakecap-manifest.tsv are intentionally committed
+# (regenerate locally after BST element changes: python3 scripts/gen-filemap.py)
 files/xattr-db/
 files/fakecap/fakecap-restore
 files/fakecap/fakecap.so
--- a/Justfile
+++ b/Justfile
@@ -109,7 +109,7 @@ export:
     $SUDO_CMD podman images | grep -E "{{image_name}}|REPOSITORY" || true
 
     # Step: Chunkify (reorganize layers)
-    # just chunkify "{{image_name}}:{{image_tag}}"
+    just chunkify "{{image_name}}:{{image_tag}}"
 
 # ── Clean ─────────────────────────────────────────────────────────────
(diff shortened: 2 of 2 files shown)


castrojo:
Pushed two fixup commits onto this branch to unblock CI:

**[fix(chunkify): pre-commit filemap, conditional regen, pin chunkah digest](https://github.com/projectbluefin/dakota/commit/c37b428)**
- Make `gen-filemap.py` conditional: skip if both `files/filemap.json` and `files/fakecap-manifest.tsv` are already present (uses `-s` test — catches missing and empty files)
- Add Justfile comment documenting regen workflow
- Pin `quay.io/coreos/chunkah` to digest `sha256:306371...` instead of `:latest`

**[chore: commit chunkah filemap and fakecap manifest for CI](https://github.com/projectbluefin/dakota/commit/02bc91d)**
- Remove the two files from `.gitignore` and commit them
- Generated from `oci/layers/bluefin.bst` deps: 840 components, 435,113 files

**Why CI was failing:** `gen-filemap.py` calls `just bst artifact list-contents` on all 840+ elements. In CI with remote execution, artifacts live on the remote CAS but not in BST's local cache — `artifact list-contents` is local-only, so it fails. Fix: treat these files like `Cargo.lock` — generated from the element tree, committed, regenerated when elements change.

**To regenerate after BST element changes:**
```bash
rm files/filemap.json files/fakecap-manifest.tsv
python3 scripts/gen-filemap.py  # requires local BST artifact cache
git add files/filemap.json files/fakecap-manifest.tsv
```

castrojo (approved):

PR #248, opened 2026-04-16
feat: add aarch64 native build

hanthor:
## Summary

- Adds a native aarch64 build job using GitHub's `ubuntu-24.04-arm` runner
- Passes `--option arch aarch64` to BuildStream — the `project.conf` already declares this arch
- GNOME's artifact cache (`gbm.gnome.org`) should have aarch64 pre-built artifacts, so most elements are cache hits
- After x86_64 + aarch64 jobs complete, a `create-manifest` job combines them into a multi-arch `ghcr.io/projectbluefin/dakota:latest`

## Motivation

Needed to run Dakota on ARM hardware, specifically the ThinkPad X13s (Qualcomm SC8280XP). An aarch64 image would allow layering X13s hardware support on top of Dakota via bootc.

## Notes

- `ubuntu-24.04-arm` runners are now generally available on GitHub Actions
- The `create-manifest` job depends on the existing x86_64 `build` job — if that job is renamed this PR needs updating
- Disk space setup mirrors the existing x86_64 job (BTRFS on extra block device)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -218,7 +218,7 @@ jobs:
         if: always()
         uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7
         with:
-          name: buildstream-logs
+          name: buildstream-logs-x86_64
           path: logs/
           retention-days: 7
           if-no-files-found: ignore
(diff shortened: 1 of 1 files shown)


castrojo:
Tested this in castrojo/dakota fork (castrojo/dakota#107). The first cold-cache aarch64 build ran for 3h03m before being cancelled — it was actively making progress (built linux-firmware, bootc, ~40 other elements) but has no CAS artifacts to pull from yet since upstream has never run an aarch64 build.

Suggestion: add `timeout-minutes: 360` to the aarch64 job so the first cold-cache run has enough time to complete:

```yaml
build-aarch64:
  runs-on: ubuntu-24.04-arm
  timeout-minutes: 360   # cold cache: full freedesktop-sdk+gnome-build-meta from scratch
  continue-on-error: true
```

Once it completes once and warms the CAS, future runs will be fast (~10-15 min like x86_64).

castrojo:
## Updated implementation

Rebased onto current `main` and rewrote to follow dakota CI best practices.

### Changes vs original

**Justfile change dropped** — the `dest` → `destination` fix is moot. Upstream rewrote `chunkify` entirely between PR open and now; it no longer uses `--mount=type=image` at all.

**ARM workflow merged into `build.yml`** (not a separate file) — cross-workflow `needs:` is not supported in GHA, so a separate `build-aarch64.yml` cannot coordinate with the existing `build` job for manifest creation. Merging both into `build.yml` allows the `create-manifest` job to properly `needs: [build, build-aarch64]`.

**Tooling aligned with existing `build` job:**
- `podman` throughout (not `docker`)
- `just bst` with `BST_FLAGS: ... --option arch aarch64` (same bst2 container, same Justfile recipes)
- Same pinned action SHAs (`actions/checkout`, `container-storage-action`, `remove-unwanted-software`, `setup-just`)
- Full `buildstream-ci.conf` with mTLS cert support for `cache.projectbluefin.io`

**ARM failures don't block x86 publication** — `continue-on-error: true` on `build-aarch64`. x86 `:latest` publishes independently. `create-manifest` only runs when ARM also succeeds.

**Artifact names made unique** — `buildstream-logs-x86_64` / `buildstream-logs-aarch64` to avoid collision within the same workflow run.

PR #251, opened 2026-04-16
chunkify: enable in export, fix CI filemap generation, pin chunkah digest

castrojo:
## Summary

Builds on #246 (hanthor's chunkify enablement) with fixes needed for CI to pass.

## Changes

**From #246:**
- Uncomment `just chunkify` in the `export` target
- Reduce `--max-layers` from 128 → 120

**CI fix:**
- Make `gen-filemap.py` conditional: skip if `files/filemap.json` and `files/fakecap-manifest.tsv` are already present (using `-s` test for non-empty files)
- Remove `files/filemap.json` and `files/fakecap-manifest.tsv` from `.gitignore` and commit them
- Add Justfile comment documenting the regen workflow

**Reproducibility:**
- Pin `quay.io/coreos/chunkah` to digest `sha256:306371251e61cc870c8546e225b13bdf2e333f79461dc5e0fc280cc170cee070` instead of `:latest`

## Why `gen-filemap.py` fails in CI

`gen-filemap.py` calls `just bst artifact list-contents` on all 840+ elements in the BST dependency tree. In CI, the build uses remote execution at `cache.projectbluefin.io`; after `just bst build`, artifacts exist in the remote CAS but **not** in BST's local cache. `artifact list-contents` is local-cache-only → fails.

Fix: treat these files like `Cargo.lock` or `go.sum` — generated from the BST element tree, committed, and regenerated when elements change. The Justfile now skips generation if both files are present.

**To regenerate after BST element changes:**
```bash
rm files/filemap.json files/fakecap-manifest.tsv
python3 scripts/gen-filemap.py  # requires local BST artifact cache
git add files/filemap.json files/fakecap-manifest.tsv
```

## Filemap stats

Generated from `oci/layers/bluefin.bst` deps: **840 components, 435,113 files**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,8 +19,8 @@
 
 # Git worktrees
 .worktrees/
-files/filemap.json
-files/fakecap-manifest.tsv
+# files/filemap.json and files/fakecap-manifest.tsv are intentionally committed
+# (regenerate locally after BST element changes: python3 scripts/gen-filemap.py)
 files/xattr-db/
 files/fakecap/fakecap-restore
 files/fakecap/fakecap.so
--- a/Justfile
+++ b/Justfile
@@ -109,7 +109,7 @@ export:
     $SUDO_CMD podman images | grep -E "{{image_name}}|REPOSITORY" || true
 
     # Step: Chunkify (reorganize layers)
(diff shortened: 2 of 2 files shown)


castrojo:
Closing — changes will be applied directly to #246 instead.

PR #258, opened 2026-04-17
ci: disable remote-execution for aarch64 build

castrojo:
## Problem

Two related bugs in the `build-aarch64` CI job:

### 1. Remote execution rejects aarch64

The BST build fails immediately with:
```
FAILURE oci/os-release.bst: Unsupported ISA "aarch64"
```
The remote execution service at `cache.projectbluefin.io:11002` only has x86_64 workers. 58 elements fail per run. Fix: remove the `remote-execution:` block from the aarch64 CI config so BST builds locally on `ubuntu-24.04-arm`. Artifacts/source-caches still use the CAS server.

### 2. `create-manifest` gates on wrong condition

GitHub Actions documents that `needs.<job>.result` returns `'success'` for jobs with `continue-on-error: true`, **even when the job fails**. So the existing condition:
```yaml
needs.build-aarch64.result == 'success'
```
was always true, causing `create-manifest` to run after every publish build even when arm64 never pushed `:aarch64`. This could corrupt `:latest` by pushing a manifest that references a non-existent or stale aarch64 image.

Fix: add `pushed=true` step output to the aarch64 Push to GHCR step. Gate `create-manifest` on `needs.build-aarch64.outputs.pushed == 'true'` — only set when tags are actually in the registry.

## Fixes

1. Remove `remote-execution:` from aarch64 BST config (build locally on arm runner)
2. Gate `create-manifest` on step output, not job result

## Validation

- `feat/aarch64-build` (SHA `ecd36d7e`) succeeded with local builds in 2h30m — proves disk/time is fine
- Step output pattern ensures `create-manifest` only runs when `:aarch64` is actually in the registry

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -357,27 +357,10 @@ jobs:
                 client-key: /src/client.key
                 client-cert: /src/client.crt
 
-          remote-execution:
-            execution-service:
-              url: https://cache.projectbluefin.io:11002
-              connection-config:
-                keepalive-time: 60
-                retry-limit: 5
-                retry-delay: 1000
-                request-timeout: 180
-              auth:
-                client-key: /src/client.key
-                client-cert: /src/client.crt
-            action-cache-service:
-              url: https://cache.projectbluefin.io:11002
-              connection-config:
-                keepalive-time: 60
-                retry-limit: 5
-                retry-delay: 1000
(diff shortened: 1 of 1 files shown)


castrojo:
Closing to avoid triggering CI and cancelling in-progress production build. Will reopen when safe.

PR #261, opened 2026-04-18
ci(aarch64): add native aarch64 build and multi-arch manifest

castrojo:
Adds a parallel native aarch64 build job on `ubuntu-24.04-arm` runners and a `create-manifest` job that combines both arches into a multi-arch `:latest` manifest.

## Fixes
- aarch64 Login/Tag/Push steps were gated on `event_name == push && ref == refs/heads/main` — a trigger that no longer exists. Fixed to use `merge_group | schedule | workflow_dispatch` consistent with the amd64 job.
- `create-manifest` was missing the event type guard, causing it to run on `pull_request` events before either arch image was pushed to GHCR (`manifest unknown` error). Fixed.

## Behavior
- ARM failures use `continue-on-error: true` and gate on a job output (`build_ok`) — arm failures do **not** block x86_64 publication
- `create-manifest` only runs on publish events when both arches succeeded

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -435,8 +435,9 @@ jobs:
       - name: Login to GHCR
         if: >-
           steps.arm_build.outcome == 'success' &&
-          github.event_name == 'push' &&
-          github.ref == 'refs/heads/main'
+          (github.event_name == 'merge_group' ||
+           github.event_name == 'schedule' ||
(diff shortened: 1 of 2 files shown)


castrojo:
Superseded by already-merged PRs #248 (native aarch64 build job) and #258 (remote exec fix). This branch is stale and diverged from upstream/main. Closing — no changes needed.

PR #276, opened 2026-04-19
fix: one-time migration for LUKS BLS entries to use rd.luks.name for DPS boot compatibility

### LINKED ISSUE #275: bootc switch fails to boot on LUKS systems: missing root= after switch due to DPS expectation
## Problem

After `bootc switch` (or `bootc upgrade`) on systems installed with the tuna/fisherman installer, the switched-to image fails to boot and the system auto-rolls back.

## Root Cause

**bootc 1.15.1** intentionally strips `root=UUID=<uuid>` from the generated BLS entry ([source](https://github.com/containers/bootc/blob/v1.15.1/crates/lib/src/bootc_composefs/boot.rs#L603-L606)):

```rust
// Remove "root=" from kernel cmdline as systemd-auto-gpt-generator should use DPS
cmdline_refs.remove(&ParameterKey::from("root"));
```

This is by design — bootc expects the system to use **Discoverable Partition Specifications (DPS)** via `systemd-gpt-auto-generator`, where the LUKS partition has GPT type `CA7D7CCB-63ED-4C53-861C-1742536059CC` (Linux LUKS) so systemd can auto-discover and unlock the root after boot.

**However**, the fisherman installer creates the LUKS data partition with GPT type `Linux filesystem` (`0FC63DAF-...`) instead of `Linux LUKS` (`CA7D7CCB-...`). Without the correct type, `systemd-gpt-auto-generator` cannot auto-discover the LUKS root, and dracut drops to emergency shell because it cannot find `/sysroot`.

## Symptoms

- `bootc switch` succeeds and stages the new image
- On next boot, system appears to boot but either hangs at LUKS prompt or drops to emergency shell (no journal captured — fails before journald starts)
- System auto-rolls back to the previous image
- `bootc status` shows `bootOrder: rollback` and previous image as `booted`
- If the new BLS entry is manually patched to add back `root=UUID=<uuid>`, the boot succeeds

## Affected systems

All systems installed with the fisherman/tuna-installer prior to a fix being applied to set the correct LUKS GPT partition type. Confirmed on a real x86_64 install with LUKS+btrfs+composefs.

## Fix

Two changes needed:

1. **fisherman installer** (see tuna-os/fisherman#19 for the EFI size fix — the LUKS partition type fix should also go there): set the data partition GPT type to `CA7D7CCB-63ED-4C53-861C-1742536059CC` when creating a LUKS-encrypted install, so DPS works on new installs.

2. **Dakota** (this repo): a one-time migration systemd service that detects existing installs where the LUKS partition has the wrong GPT type and corrects it using `sgdisk`. This fixes upgrades on existing systems without reinstall.

## Workaround for existing systems

Until fixed, manually add `root=UUID=<uuid>` back to the staged BLS entry after `bootc switch`:

```bash
# Get root UUID
ROOT_UUID=$(findmnt -n -o UUID /)

# Find and patch the staged BLS entry
sudo mount /dev/disk/by-partlabel/EFI /efi 2>/dev/null || true
CONF=$(grep -rl 'sort-key bootc.*-0' /efi/loader/entries/*.conf | head -1)
sudo sed -i "s|options rootflags=|options root=UUID=${ROOT_UUID} rootflags=|" "$CONF"
```
Then reboot. The boot will still prompt for LUKS passphrase (TPM2 re-enrollment needed separately).

hanthor:
## Problem

**bootc >= 1.15.1** strips `root=UUID=` from BLS boot entries and relies on `systemd-gpt-auto-generator` (DPS) to find the root. DPS requires the LUKS container to be mapped as `/dev/mapper/root`.

The older fisherman installer wrote `rd.luks.uuid=<UUID>` into BLS entries, which maps LUKS to `/dev/mapper/luks-<UUID>` — not `/dev/mapper/root`. So `bootc switch`/`bootc upgrade` generates an unbootable BLS entry and the system auto-rolls back.

**fisherman PR #18** (merged 2026-04-19) fixed this for **new installs** by switching to `rd.luks.name=<UUID>=root`. This PR provides the **migration for existing installs**.

Closes #275

## Fix

Adds `elements/bluefin/luks-gpt-migration.bst` which installs a one-time systemd service (`bluefin-fix-luks-boot.service`) that:

1. **Updates all BLS entries on the ESP** — replaces `rd.luks.uuid=<UUID>` with `rd.luks.name=<UUID>=root` so the LUKS maps to `/dev/mapper/root` as DPS requires. `bootc upgrade` then inherits the fixed cmdline into new entries.

2. **Corrects the LUKS partition GPT type** — changes `0FC63DAF-...` (Linux filesystem) to `CA7D7CCB-...` (Linux LUKS) as defense-in-depth for `systemd-gpt-auto-generator` partition discovery.

The script uses `rd.luks.uuid=` from `/proc/cmdline` to identify exactly the boot LUKS partition, avoiding any risk of touching secondary encrypted drives.

## TPM2 note

Fisherman enrolls **PCR 7 only** (Secure Boot state). Since PCR 7 does not measure the kernel cmdline, patching BLS entries **does not affect TPM2 auto-unlock** — no re-enrollment needed.

### DIFF
--- a/elements/bluefin/deps.bst
+++ b/elements/bluefin/deps.bst
@@ -28,6 +28,7 @@ depends:
   - bluefin/sudo-rs.bst
   - bluefin/efibootmgr.bst
   - bluefin/xdg-terminal-exec.bst
+  - bluefin/luks-gpt-migration.bst
 
   # Include things missing from the gnomeos base image
   - freedesktop-sdk.bst:components/buildstream2.bst
--- a/elements/bluefin/luks-gpt-migration.bst
+++ b/elements/bluefin/luks-gpt-migration.bst
@@ -1,0 +1,143 @@
+kind: manual
+
+# No external sources - all files are inline.
+build-depends:
+- freedesktop-sdk.bst:public-stacks/runtime-minimal.bst
+
+config:
+  install-commands:
+  - |
+    install -Dm755 /dev/stdin "%{install-root}%{indep-libdir}/bluefin/fix-luks-boot" <<'SCRIPT'
+    #!/bin/bash
+    # One-time migration for systems installed with older fisherman builds that
+    # wrote rd.luks.uuid=<UUID> into BLS entries instead of rd.luks.name=<UUID>=root.
+    #
+    # bootc >= 1.15.1 strips root=UUID= from generated BLS entries and relies on
+    # systemd-gpt-auto-generator (DPS) to find the root. DPS requires the LUKS
+    # container to be mapped as /dev/mapper/root. rd.luks.uuid= maps it to
+    # /dev/mapper/luks-<UUID> instead, so DPS fails and the system won't boot after
+    # bootc switch/upgrade.
+    #
+    # This script:
+    #   1. Updates all BLS entries on the ESP to use rd.luks.name=<UUID>=root
+    #      (mirrors what fisherman >= 2026-04-19 does for new installs)
+    #   2. Also corrects the LUKS partition GPT type to CA7D7CCB-... (Linux LUKS)
+    #      as defense-in-depth (older fisherman used the generic Linux filesystem type)
+    #
+    # NOTE: Updating BLS entries changes the cmdline measured by TPM2 PCR8.
+    # If TPM2 auto-unlock is enrolled, you will need to re-enroll after this runs.
+    #
+    # References:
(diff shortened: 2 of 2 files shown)


castrojo:
So how does this work, if I have a busted one I boot into the known good image and then when this update comes that's it?

PR #314, opened 2026-04-22
linux: enable IPV6_MULTIPLE_TABLES for tailscaled IPv6 routing

hanthor:
## Problem

Without `CONFIG_IPV6_MULTIPLE_TABLES`, `ip -6 rule` fails entirely:

```
$ ip -6 rule
RTNETLINK answers: Address family not supported by protocol
```

This causes `tailscaled` to log:

```
router: disabling tunneled IPv6 due to system IPv6 config:
  kernel doesn't support IPv6 policy routing
```

Without the ability to install IPv6 routing rules, tailscaled falls back to using `fe80::` link-local IPv6 paths when reaching LAN peers on the tailnet. These link-local paths are sticky and don't recover after sleep/wake cycles — disco pings continue to work (they use a different code path) while all WireGuard data packets time out silently.

The Fedora kernel has had `IPV6_MULTIPLE_TABLES` enabled for years; this is a straightforward gap in the Dakota kernel config.

## Symptoms

- `tailscale ping <lan-peer>` (disco) succeeds
- `tailscale ping --tsmp <lan-peer>` times out
- `tailscale status` shows peer at `[fe80::...%iface]:41641`
- Rebooting or restarting `tailscaled` temporarily fixes it

## Fix

Enable `IPV6_MULTIPLE_TABLES` and `IPV6_SUBTREES` in `fdsdk-config.sh`.

`IPV6_SUBTREES` enables source-address-based routing and is the natural companion to `IPV6_MULTIPLE_TABLES`; Fedora enables both.

## Testing

Verified on a Dakota alpha install (dilli, `ghcr.io/projectbluefin/dakota`):
- Before: `ip -6 rule` → `Address family not supported by protocol`
- After enabling the module manually: tailscaled installs its fwmark routing rules and uses stable IPv4/global-IPv6 paths to LAN peers instead of link-local

Closes #174 (follow-up to #175)


hanthor:
Closing — PR branch was based on a stale fork that included unrelated commits. Re-opening with a minimal diff.

PR #357, opened 2026-04-29
fix(deps): use codeberg closer-mop repo

AlexNPavel:
User pcostanza deleted their GitHub account and the repo is now only available on codeberg.


castrojo (approved):
Thanks!

PR #374, opened 2026-05-01
fix: Remove passwordless sudo policy

### LINKED ISSUE #361: Default allow `sudo` without password
By default local user can use `sudo` without password. There's a `/etc/sudoers.d/wheel-nopasswd` which contains:

```
%wheel ALL=(ALL) NOPASSWD: ALL
```

I assume this is from GNOME OS? Unless this is intended maybe we should remove it.

coxde:
## Description

Remove the inherited GNOME OS passwordless sudo drop-in from Dakota’s shared OCI stack.

GNOME OS currently contributes `/etc/sudoers.d/wheel-nopasswd`, which makes `sudo` passwordless for members of `wheel` by default. Dakota should not ship that as the default behavior.

This removes only the `wheel-nopasswd` drop-in during image integration, while leaving the normal passworded `wheel` sudo policy intact.

Close #361

## Changes

- **`elements/oci/layers/bluefin-stack.bst`**
  - remove `/etc/sudoers.d/wheel-nopasswd` during shared OCI stack integration
  - add a short comment explaining that this overrides inherited GNOME OS behavior

## Type of Change

- [x] Bug fix

## Testing

- [x] `just build`
- [x] VM boot test

Assisted-by: GPT-5.4 via Zed

## Summary by CodeRabbit

* **Chores**
  * System sudo authentication configuration has been updated for the wheel group. Users will now be required to enter their password when executing privileged commands via sudo, replacing the previous passwordless policy. This change applies to all members of the wheel group and affects how privilege escalation operations are authenticated on the system.


castrojo:
This looks fine, feel free to undraft when ready, thanks!

castrojo (approved):

PR #384, opened 2026-05-02
ci: move export/chunkify/sbom to build, publish loads artifact

castrojo:
Build now owns the full OCI pipeline while BST artifacts are hot in local cache: export (includes chunkah rechunking into 120 plain-zstd layers), bootc container lint, SBOM generation, and podman save to a BTRFS-backed path. The post-chunkify OCI archive is uploaded as a GitHub artifact (3-day retention, default variant only, non-PR events).

Publish is stripped to: download artifact, podman load, push, sign. No BST, no CAS download, no chunkify in publish. Blacksmith is still used for non-schedule runs for GHCR push throughput.

Key correctness constraints honoured:
- container-storage-action + remove-unwanted-software run before any podman call in build (BST uses rootful podman for bst2 image pull)
- chunkify runs exactly once (inside just export) — no double-rechunk
- plain podman push --compression-format=zstd (not skopeo, not zstd:chunked)
- actions: read permission added for cross-run artifact download
- workflow_dispatch gains build_run_id/build_sha inputs with null guard for the auto-detect path (fail fast, not silent hang)
- artifact upload gated to matrix.variant == 'default' only
- upload-artifact pin updated to v7.0.1 (043fb46d)

Assisted-by: Claude Sonnet 4.6 via GitHub Copilot

## Summary by CodeRabbit

* **Chores**
  - Optimized CI build pipeline with improved storage preparation and artifact handling.
  - Pinned and updated CI actions to more stable versions.
* **New Features**
  - Publish workflow accepts optional manual inputs for targeting specific builds.
  - Build artifacts now include an OCI image export and SBOM that are uploaded and reused for publishing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -85,7 +85,7 @@ jobs:
       # Both steps must run before any podman call (BST uses rootful podman
       # to pull bst2; storage backend must be stable before first use).
       - name: Remove unwanted software
-        uses: ublue-os/remove-unwanted-software@695eb75bc387dbcd9685a8e72d23439d8686cba6 # v10
+        uses: ublue-os/remove-unwanted-software@cc0becac701cf642c8f0a6613bbdaf5dc36b259e # v9
 
       - name: Mount BTRFS for podman storage
         id: container-storage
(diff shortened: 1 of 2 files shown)


castrojo:
Addressing CodeRabbit's two comments:

**`continue-on-error` on container-storage (build.yml):** Intentional. Blacksmith runners have no `/mnt` BTRFS device, so this action silently no-ops and podman falls back to the root FS (~50GB free after `remove-unwanted-software`). Removing this would hard-fail every Blacksmith run before any podman work starts. This is the designed behavior for this runner environment.

**Branch validation on manual dispatch (publish.yml):** Valid security concern — tracking as a follow-up. Will add `head_branch` guard to the `gh api` lookup to ensure manual dispatches only publish from `main` or `gh-readonly-queue/main/*`.

Merging now under admin authority to unblock today's publish.

Issue #331, opened 2026-04-25
Menu option "Export Command List" does not work


StorageB:
In the Custom Command Menu extension preferences, the Export Command List option does not work.

As mentioned [here](https://github.com/projectbluefin/dakota/issues/228#issuecomment-4310405377), the extension expects `command-order` to be 1-99 (number of entries the extension allows) instead of 1-11. The export fails after the 11th entry as it is searching through all available keys. This should be more robust in the upstream code, but changing `command-order` should fix this for now.

PR #416, opened 2026-05-09
Remove command-order from 05-dakota-custom-command-menu

StorageB:
This commit https://github.com/projectbluefin/dakota/commit/031b62f90af7cec1574d2b17cea899147e9de4f0 reset `command-order` back to 1-11 reintroducing this issue https://github.com/projectbluefin/dakota/issues/331 where the extension expects `command-order` to contain 99 values (the number of entries allowed by the extension).

Instead of adding 1-99 back to this file, I believe we can just remove `command-order` entirely and allow it to use the default schema values where it is defined as 1-99.

## Summary by CodeRabbit

* **Chores**
  * Streamlined the custom command menu configuration by removing redundant ordering entries while preserving behavior.

* **Documentation**
  * Updated the README’s ISO Archive table: removed older entries and added new ISO listings (dated early May 2026) with updated checksum links.

[Review Change Stack](https://app.coderabbit.ai/change-stack/projectbluefin/dakota/pull/416)

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,13 +27,20 @@ Some ISOs are duds, so here's an archive:
 <!-- iso-table-start -->
 | Date | ISO | Checksum |
 |------|-----|----------|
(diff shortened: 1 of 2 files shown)


castrojo:
Thanks for catching this I tested the config locally and it was working fine before the 99 things. Merging now thanks!

PR #446, opened 2026-05-15
bluefin: disable shell motd by default

hanthor:
## Summary
- set BLUEFIN_SHELL_ENABLE_MOTD=0 by default in the built image
- write it into /etc/environment from elements/bluefin/common.bst
- remove any pre-existing BLUEFIN_SHELL_ENABLE_MOTD entry before appending

## Why
This makes the Bluefin shell MOTD disabled by default in Dakota images.

## Summary by CodeRabbit

* **Chores**
  * Adds a shell profile defaults script that disables the Bluefin shell MOTD by default.
  * The setting is exported and can be overridden by users if desired.

[Review Change Stack](https://app.coderabbit.ai/change-stack/projectbluefin/dakota/pull/446)

### DIFF
--- a/elements/bluefin/common.bst
+++ b/elements/bluefin/common.bst
@@ -42,6 +42,12 @@ config:
     cp -r system_files/shared/etc/./ "%{install-root}%{sysconfdir}/"
     cp -r bluefin-branding/system_files/etc/./ "%{install-root}%{sysconfdir}/"
 
+    # Disable Bluefin shell MOTD by default.
+    ENV_FILE="%{install-root}%{sysconfdir}/environment"
+    touch "${ENV_FILE}"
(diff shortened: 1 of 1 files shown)


castrojo:
I don't understand this change?

repires:
> I don't understand this change?

Removes the doubled motd message, it will come from common only.

hanthor:
> I don't understand this change?

Since we are shipping the bluefin-cli program in Dakota it means that the more get doubled since it's coming from bluefin-cli and common. This tells bluefin-cli to default to off

castrojo:
Oh I thought we replicated bluefin cli in here not use bluefin cli itself. I'm confused by the brew package. 

(This should probably be a seperate issue but should we think about just dropping bluefin cli and just building it in? Does it make sense? I'm asking because I want bluefin-cli but if I install the brewfile it'll pull in things that are already on the image.)

castrojo (approved):

PR #573, opened 2026-05-28
Fix conditional check for BUILD_SKIP_CHUNKIFY

hanthor:

---

## Testing

- [x] `just validate` passes
- [x] `just boot-test` passes (automated boot smoke test)
- [x] `just lint` passes on a built image

## Checklist

````markdown
```verify
```
````

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration handling to enforce stricter validation of required build settings. The build process will now error if expected variables are not explicitly set, rather than treating them as empty values.

[Review Change Stack](https://app.coderabbit.ai/change-stack/projectbluefin/dakota/pull/573?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -527,7 +527,7 @@ chunkify image_ref:
     #!/usr/bin/env bash
     set -euo pipefail
 
-    if [ "${BUILD_SKIP_CHUNKIFY:-}" = "1" ]; then
+    if [ "$BUILD_SKIP_CHUNKIFY" = "1" ]; then
         echo "==> Skipping chunkify (BUILD_SKIP_CHUNKIFY=1)"
         exit 0
     fi


castrojo:
This change introduces a regression. Removing the `:-` default operator from `${BUILD_SKIP_CHUNKIFY:-}` will cause an unbound variable error when `set -euo pipefail` is active and `BUILD_SKIP_CHUNKIFY` is not set — which is the normal case for most developers running `just chunkify`.

The current pattern `${BUILD_SKIP_CHUNKIFY:-}` is correct: it expands to an empty string when the variable is unset, allowing the comparison to `"1"` to evaluate safely.

Compare with the existing pattern at line 92: `${BUILD_SKIP_NVIDIA:-}` — same idiom, correct for the same reason.

Recommend closing this PR. If stricter variable enforcement is wanted, the right approach would be to set an explicit default at the top of the recipe (e.g. `BUILD_SKIP_CHUNKIFY=${BUILD_SKIP_CHUNKIFY:-0}`), not remove the default expansion.

PR #584, opened 2026-05-28
Remove command-order from 05-dakota-custom-command-menu

StorageB:
`command-order` should not be defined here because it reintroduces two separate bugs. It was previously removed here https://github.com/projectbluefin/dakota/pull/416 but was added back by this commit https://github.com/projectbluefin/dakota/commit/fa4bd1441bb7a36b6ddbce109f608a7516827f82

## Summary by CodeRabbit

* **Chores**
  * Simplified custom command menu configuration by removing explicit ordering definition. Menu commands now rely on sequential command definitions instead.

[Review Change Stack](https://app.coderabbit.ai/change-stack/projectbluefin/dakota/pull/584?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/files/dconf/05-dakota-custom-command-menu
+++ b/files/dconf/05-dakota-custom-command-menu
@@ -14,7 +14,6 @@ menuoptions-setting=2
 menuicon-setting='ublue-logo-symbolic'
 menulocation-setting=1
 menuposition-setting=0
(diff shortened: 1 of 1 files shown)


castrojo (approved):
Clean revert of a regression. `command-order` was correctly removed in #416 and shouldn't have come back. No lab run needed — dconf config only.

PR #610, opened 2026-05-30
feat(ujust): add cleanup-stale-boot-entries recipe (#531)

### LINKED ISSUE #531: bootc status fails with 'Multiple extra entries in /boot' after image rename
# Bug Report: `bootc status` fails with "Multiple extra entries in /boot, could not determine rollback entry" after OS/image rename

## Description
When running `sudo bootc status --format json` on a system configured with the `composefs` backend, the command fails with the following error:
```
error: Status: Getting composefs deployment status: Getting composefs deployment status: Multiple extra entries in /boot, could not determine rollback entry
```

This error is encountered on **Bluefin Dakota** (which uses a pure `bootc` and `composefs` backend without OSTree). 

---

### 1. The Duplicate Bootloader Configurations
Upon inspecting the boot entries in `/boot/loader/entries/`, we find four `.conf` files:
```
/boot/loader/entries/bootc_bootc-20260507-1.conf
/boot/loader/entries/bootc_bootc-20260516-0.conf
/boot/loader/entries/bootc_bluefin_dakota-20260516-0.conf
/boot/loader/entries/bootc_bluefin_dakota-20260520-1.conf
```

Inside the configuration files, both `bootc_bootc-20260516-0.conf` and `bootc_bluefin_dakota-20260516-0.conf` reference the **exact same** composefs verity digest:
* **`bootc_bluefin_dakota-20260516-0.conf`:**
  ```ini
  title Bluefin
  version 20260516
  linux /EFI/Linux/bootc_composefs-230a0ec9440a21baea5e14a15135e8396cbd7ea225a40162e4d2b9ad451b3551e2e7d334681dcf4ba49e86f7cad8d9966cbac8928ca9e3c4d48ab5f4c1c8cb58/vmlinuz
  initrd /EFI/Linux/bootc_composefs-230a0ec9440a21baea5e14a15135e8396cbd7ea225a40162e4d2b9ad451b3551e2e7d334681dcf4ba49e86f7cad8d9966cbac8928ca9e3c4d48ab5f4c1c8cb58/initrd
  options root=UUID=a906ce76-f256-4ba1-a260-b6b12da7fe1a rootflags=subvol=/ rw composefs=b6fa523f25053ee1f42651e1214e89d2a19c47ee26222b6fe3955914a35a9340d1e573d92c64ddf82002c9544b7e5d4d3a56a409f3e1ba031b7f429c832d4fd9 rhgb quiet rd.luks.name=3486c9a9-84b8-4448-a733-6c6a48ce2444=root
  sort-key bootc-bluefin-dakota-1
  ```
* **`bootc_bootc-20260516-0.conf`:**
  ```ini
  title Bluefin
  version 20260516
  linux /EFI/Linux/bootc_composefs-230a0ec9440a21baea5e14a15135e8396cbd7ea225a40162e4d2b9ad451b3551e2e7d334681dcf4ba49e86f7cad8d9966cbac8928ca9e3c4d48ab5f4c1c8cb58/vmlinuz
  initrd /EFI/Linux/bootc_composefs-230a0ec9440a21baea5e14a15135e8396cbd7ea225a40162e4d2b9ad451b3551e2e7d334681dcf4ba49e86f7cad8d9966cbac8928ca9e3c4d48ab5f4c1c8cb58/initrd
  options root=UUID=a906ce76-f256-4ba1-a260-b6b12da7fe1a rootflags=subvol=/ rw composefs=b6fa523f25053ee1f42651e1214e89d2a19c47ee26222b6fe3955914a35a9340d1e573d92c64ddf82002c9544b7e5d4d3a56a409f3e1ba031b7f429c832d4fd9 rhgb quiet rd.luks.name=3486c9a9-84b8-4448-a733-6c6a48ce2444=root
  sort-key bootc-bootc-1
  ```

This state occurred because the system transitioned from using the generic `bootc` image/OS name prefix to the tailored `bootc_bluefin_dakota` prefix. 
* During updates, `bootc`'s garbage collection and prune logic only manages boot configuration files matching the current prefix (`bootc_bluefin_dakota`), thereby leaving the older `bootc_bootc-*.conf` files orphaned in `/boot/loader/entries/`.
* However, when reading boot entries to compute deployment status, `bootc` scans *all* files ending in `.conf` under the entries directory, which includes the stale files.

---

### 2. Deep Dive Into Upstream `containers/bootc` Code

The error originates in `crates/lib/src/bootc_composefs/status.rs` within the `composefs_deployment_status_from` function:

1. **Listing Bootloader Entries:**
   The function retrieves all `.conf` boot configurations:
   ```rust
   let bootloader_entry_verity = list_bootloader_entries(storage)?;
   ```
   Because there are two files with the same verity digest (`b6fa5...`), `bootloader_entry_verity` contains two entries with the exact same `fsverity` value.

2. **Categorizing Deployments:**
   The code iterates through all entries to construct deployment lists:
   ```rust
   for BootloaderEntry {
       fsverity: verity_digest,
       ..
   } in bootloader_entry_verity
   {
       ...
       if verity_digest == booted_composefs_digest.as_ref() {
           host.spec.image = boot_entry.image.as_ref().map(|x| x.image.clone());
           host.status.booted = Some(boot_entry);
           continue;
       }

       if let Some(staged_deployment) = &staged_deployment {
           let staged_depl = serde_json::from_str::<StagedDeployment>(&staged_deployment)?;

           if verity_digest == staged_depl.depl_id {
               boot_entry.download_only = staged_depl.finalization_locked;
               host.status.staged = Some(boot_entry);
               continue;
           }
       }

       extra_deployment_boot_entries.push(boot_entry);
   }
   ```
   Since the duplicate verity digest (`b6fa5...`) is neither the booted nor the staged deployment, it gets pushed into `extra_deployment_boot_entries` **twice**.

3. **Filtering Rollback Candidates:**
   Further down, the code determines the rollback candidate by filtering `extra_deployment_boot_entries` against the set of active bootloader verity digests:
   ```rust
   let rollback_candidates: Vec<_> = extra_deployment_boot_entries
       .into_iter()
       .filter(|entry| {
           let verity = &entry
               .composefs
               .as_ref()
               .expect("composefs is always Some for composefs deployments")
               .verity;
           bootloader_configured_verity.contains(verity)
       })
       .collect();
   ```
   Both duplicate elements match the filter since `bootloader_configured_verity` contains `b6fa5...`. Thus, `rollback_candidates` ends up with a length of `2` (both containing the `b6fa5...` deployment information).

4. **The Failure Point:**
   Finally, the code runs a strict count check:
   ```rust
   if rollback_candidates.len() > 1 {
       anyhow::bail!("Multiple extra entries in /boot, could not determine rollback entry");
   }
   ```
   Because `rollback_candidates.len() == 2`, this sanity check fails and throws the error, preventing users from checking status or finalizing deployments.

---

### A. Upstream Fix (containers/bootc)
The upstream `bootc` status code should be resilient to duplicate bootloader configurations pointing to the same physical deployment. This can be solved by **deduplicating `bootloader_entry_verity` by `fsverity` digest** before processing the loop.

Inside `composefs_deployment_status_from`:
```rust
    // This is our source of truth
    let mut bootloader_entry_verity = list_bootloader_entries(storage)?;
    
    // Deduplicate entries by fsverity digest to avoid duplicate processing of the same deployment
    let mut seen = std::collections::HashSet::new();
    bootloader_entry_verity.retain(|entry| seen.insert(entry.fsverity.clone()));
```

### B. Downstream Workaround (Project Bluefin / Dakota)
To unblock affected users, they can simply manually remove the stale or orphaned `bootc_bootc-*.conf` configuration files from the EFI/boot partition:
```bash
# 1. Mount ESP to /boot if not already mounted
sudo mount /dev/nvme0n1p1 /boot

# 2. Safely remove the orphaned configs matching the old prefix
sudo rm /boot/loader/entries/bootc_bootc-*.conf
```
This instantly resolves the error and allows `bootc status` to complete successfully.

hanthor:
## Summary

Adds a `cleanup-stale-boot-entries` ujust recipe to detect and remove orphaned bootloader entries left behind after an OS/image name change (e.g. `bootc` → `bootc_bluefin_dakota`).

## Problem

When Dakota transitions between image prefixes, `bootc`'s prune logic only manages entries matching the current prefix. Old bootloader entries remain in `/boot/loader/entries/` and cause:

```
error: Multiple extra entries in /boot, could not determine rollback entry
```

## Fix

The recipe:
1. Reads the current bootc prefix from `/proc/cmdline`
2. Finds `.conf` entries under `/boot/loader/entries/` not matching the current prefix
3. Shows the user which entries are stale
4. Asks for confirmation before removing them

## Test plan

- [ ] Verify recipe appears in `ujust --list`  
- [ ] Run on a system with stale entries → entries detected and removed  
- [ ] Run on a clean system → "No stale entries found"

Closes #531

## Summary by CodeRabbit

* **New Features**
  * Added a maintenance utility to detect and remove stale boot loader entries; it lists candidates and asks for confirmation before deletion.
* **Behavior Changes**
  * Improved reboot messaging and now prompts users to reboot immediately when required.
  * Prevents direct root execution of the flow and informs users when elevated invocation is not allowed.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -27,7 +27,7 @@
         '/elements/plugins/.*\\.bst$/',
       ],
       matchStrings: [
-        'url:\\s*pypi:[a-f0-9/]+/(?<depName>buildstream[-_]plugins(?:[-_]community)?)-(?<currentValue>\\d+\\.\\d+\\.\\d+)\\.tar\\.gz\\n\\s+ref:\\s*(?<currentDigest>[a-f0-9]{64})',
+        'url:\\s*pypi:source/[a-z]/[\\w-]+/(?<depName>[\\w_-]+)-(?<currentValue>\\d+\\.\\d+\\.\\d+)\\.tar\\.gz\\n\\s+ref:\\s*(?<currentDigest>[a-f0-9]{64})',
       ],
       datasourceTemplate: 'pypi',
       versioningTemplate: 'pep440',
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -27,7 +27,7 @@
         '/elements/plugins/.*\\.bst$/',
       ],
       matchStrings: [
-        'url:\\s*pypi:source/[a-z]/[\\w-]+/(?<depName>[\\w_-]+)-(?<currentValue>\\d+\\.\\d+\\.\\d+)\\.tar\\.gz\\n\\s+ref:\\s*(?<currentDigest>[a-f0-9]{64})',
+        'url:\\s*pypi:source/[a-z0-9]/[\\w-]+/(?<depName>[\\w_-]+)-(?<currentValue>\\d+\\.\\d+\\.\\d+)\\.tar\\.gz\\n\\s+ref:\\s*(?<currentDigest>[a-f0-9]{64})',
       ],
       datasourceTemplate: 'pypi',
       versioningTemplate: 'pep440',
--- a/.github/skills/lab-commands.md
+++ b/.github/skills/lab-commands.md
@@ -1,0 +1,91 @@
+# Lab Slash Commands
+
+Lab commands control hardware validation for PRs. Only humans may issue them.
+
+## The one rule agents must never break
+
+> **Agents MUST NOT post `/lab` commands.** Not on behalf of a user. Not autonomously.
+> Ever. Only a human with write+ access types a `/lab` command.
+
+A question in a PR comment — "does this work?", "any lab results?" — is never a
+command. Questions do not trigger lab actions. If an agent is unsure whether a
+human issued a command, it does nothing and waits.
+
+---
+
+## Who can use lab commands
+
+**Maintainers** (`@projectbluefin/maintainers`) and **wranglers** (write+ collaborators).
+
+Anyone else: the command is silently ignored. No error comment, no action.
+
+---
+
+## Slash commands
+
+Commands must appear on their own line, starting with `/lab`. Nothing else on
+that line counts as a command.
+
+| Command | Effect | Label applied |
+|---------|--------|---------------|
+| `/lab run` | Dispatch a new lab workflow for this PR | `lab:pending` → `lab:running` |
+| `/lab reset` | Cancel any running workflow, dispatch a fresh run | `lab:pending` |
+| `/lab skip <reason>` | Skip lab for this PR; reason required | `lab:skip` |
+| `/lab pass` | Manual override: mark this PR lab-passed | `lab:pass` |
+| `/lab fail` | Manual override: mark this PR lab-failed | `lab:fail` |
+
+`<reason>` is mandatory for `/lab skip`. A skip without a reason is rejected.
+
+---
+
+## Label states
+
+| Label | Meaning |
+|-------|---------|
+| `lab:pending` | Run queued, not yet dispatched |
+| `lab:running` | Argo workflow in progress |
+| `lab:pass` | Lab validated — image booted on hardware |
+| `lab:fail` | Lab failed — see `<!-- lab-status -->` sentinel for details |
+| `lab:skip` | Skipped with maintainer sign-off |
+
+Labels form a state machine: `pending → running → pass|fail`. Any transition
+backwards requires an explicit `/lab reset` or `/lab run`.
+
+---
+
+## Sentinel comment
+
+Every PR that enters the lab pipeline has exactly one `<!-- lab-status -->`
+sentinel comment. Agents that update lab state **PATCH that comment in-place**
+via `gh api --method PATCH`. They never post a new lab comment. If the sentinel
+does not exist, create it once — then only PATCH from that point on.
+
+```bash
+# PATCH existing sentinel — never POST a new one
+COMMENT_ID=$(gh api repos/projectbluefin/dakota/issues/${PR}/comments \
+  | python3 -c "import json,sys; cs=[c for c in json.load(sys.stdin) if '<!-- lab-status -->' in c['body']]; print(cs[0]['id'] if cs else '')")
+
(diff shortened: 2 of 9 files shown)


castrojo (changes requested):
The recipe addresses a real problem but has a logic bug in the prefix matching:

```bash
CURRENT_PREFIX=$(grep -oP 'bootc[^\s]*' /proc/cmdline | head -1 || echo "")
# ...
if [[ "$basename" != bootc_${CURRENT_PREFIX}* ]]; then
```

If `CURRENT_PREFIX` extracts a value like `bootc_bluefin_dakota` from cmdline, the filter becomes `bootc_bootc_bluefin_dakota*` — which never matches anything, and ALL entries would be flagged as stale.

The cmdline parsing is also fragile; the bootc entry name format doesn't reliably appear verbatim in `/proc/cmdline`.

Suggested approach: derive the current prefix from `bootc status --json` instead:
```bash
CURRENT_IMAGE=$(sudo bootc status --json 2>/dev/null | jq -r '.status.booted.image.image.image // empty')
# Extract the image base name (e.g. "dakota" from "ghcr.io/projectbluefin/dakota:latest")
# then match against the corresponding entry prefix in /boot/loader/entries/
```
Or use `bootctl list --json=short` to get the active entry ID directly.

castrojo (commented):
.

hanthor:
The doubled-prefix bug (`bootc_${CURRENT_PREFIX}` where CURRENT_PREFIX already starts with `bootc_`) was already fixed in this branch — current code uses `${CURRENT_PREFIX}*` directly. Added `|| true` to the bootc status pipeline so it doesn't abort under `set -euo pipefail` when bootc exits non-zero, allowing the /proc/cmdline fallback to run correctly.

castrojo:
Closing as stale — conflicting with current main. Please re-open from a fresh branch based on upstream/main if still relevant.

PR #653, opened 2026-06-01
feat(bluefin): package and add zsh

dylanmtaylor:
This PR packages `zsh` and adds it to the Dakota dependency stack.

---

## Changes

- Created new manual BuildStream element `elements/bluefin/zsh.bst` using `romkatv/zsh-bin` pre-built static binaries (`v6.1.1`, zsh `5.8`) for both `x86_64` and `aarch64` architectures.
- Added `bluefin/zsh.bst` to the dependency stack in `elements/bluefin/deps.bst`.
- Updated `docs/skills/overview.md` to reflect that `zsh` is now packaged in Dakota.

## Testing

- [x] `just validate` passes

## Checklist

- [x] Patches in `patches/` regenerated if junction refs changed (no junction refs changed)

## Community verification

```bash
zsh --version
```

### DIFF
--- a/docs/skills/overview.md
+++ b/docs/skills/overview.md
@@ -79,7 +79,7 @@ Production Bluefin images are **Containerfile-based overlays** — they start wi
 | gum | Y | Y |
 | fzf | Y | Y |
 | fish | N | Y |
-| zsh | N | Y |
+| zsh | Y | Y |
 | tmux | N | Y |
 | fastfetch | N | Y |
 | Starship prompt | N | Y |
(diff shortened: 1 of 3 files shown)


hanthor (approved):
LGTM — zsh packaging looks straightforward.

hanthor (changes requested):
Requesting changes because this pulls zsh from romkatv/zsh-bin, which is an unofficial third-party binary release. For a base image dependency we should use an official upstream source or build from source.

PR #654, opened 2026-06-01
feat(bluefin): package and add tmux

dylanmtaylor:
This PR packages `tmux` and adds it to the Dakota dependency stack.

---

## Changes

- Created new manual BuildStream element `elements/bluefin/tmux.bst` using `mjakob-gh/build-static-tmux` pre-built stripped static binaries (`v3.6b`) for both `x86_64` and `aarch64` architectures.
- Added `bluefin/tmux.bst` to the dependency stack in `elements/bluefin/deps.bst`.
- Updated `docs/skills/overview.md` to reflect that `tmux` is now packaged in Dakota.

## Testing

- [x] `just validate` passes

## Checklist

- [x] Patches in `patches/` regenerated if junction refs changed (no junction refs changed)

## Community verification

```bash
tmux -V
```

### DIFF
--- a/docs/skills/overview.md
+++ b/docs/skills/overview.md
@@ -80,7 +80,7 @@ Production Bluefin images are **Containerfile-based overlays** — they start wi
 | fzf | Y | Y |
 | fish | N | Y |
 | zsh | N | Y |
-| tmux | N | Y |
+| tmux | Y | Y |
 | fastfetch | N | Y |
 | Starship prompt | N | Y |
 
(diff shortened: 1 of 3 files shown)


hanthor (approved):
LGTM — tmux packaging looks straightforward.

hanthor (changes requested):
Requesting changes because this pulls tmux from mjakob-gh/build-static-tmux, an unofficial third-party binary release. For a base image dependency we should use an official upstream source or build from source.

PR #656, opened 2026-06-01
fix: remove bootc installer flatpak on first boot

### LINKED ISSUE #635: bootc installer is present on the installed system
```
Bluefin 🦖  ·  issue pipeline
─────────────────────────────────────────────────
  ✓  filed      report received
  ✓  approved   signed off by a maintainer
  ✓  queued     —
  ✓  claimed    —
  ▶  done       fix shipped
─────────────────────────────────────────────────
  report:       missing     ·  verified: 0/3
  area:         —         ·  priority: —
  next action:  ujust verify 635 — three verifies closes the case
```
🦖 rawr

---
```
DAKOTA 🦖  ·  issue pipeline
─────────────────────────────────────────────────
  ▶  filed      report received
  ·  approved   —
  ·  queued     —
  ·  claimed    —
  ·  done       —
─────────────────────────────────────────────────
  report:       missing     ·  confirms: 0
  area:         —         ·  priority: —
  next action:  same bug? ujust confirm 635
```

---### Before submitting

### ujust report gist URL

_No response_

### What happened?

I installed the system and the installed image has bootc installer available as an app. 

### Steps to reproduce

Install Dakota

### Affected area

install / ISO

### Extra context

_No response_

castrojo:
## Summary

The ISO's tuna-installer (fisherman) copies all system flatpaks to the target during installation. This includes the bootc-installer itself, leaving it visible as an app on the installed system.

## Fix

Add a firstboot oneshot service (`bluefin-remove-installer.service`) that:
1. Checks for `org.bootcinstaller.Installer` and `.Devel` variants
2. Removes them if present
3. Prunes unused runtimes
4. Writes a stamp file so it never runs again

## Checklist

- [x] Branched from `upstream/main`
- [x] `just bst show bluefin/firstboot-date.bst` passes
- [x] Minimal diff (2 files, 22 lines)

Closes #635

## Summary by CodeRabbit

* **New Features**
  * System now automatically removes installer Flatpak applications on first boot to avoid leaving installer artifacts on installed systems.
* **Documentation**
  * Added a lessons-learned entry explaining the installer leak and the defensive mitigation.
  * Expanded quickstart with an end-to-end workflow for fixing issues, branching, commit trailer rules, and PR guidance.

### DIFF
--- a/docs/skills/installer.md
+++ b/docs/skills/installer.md
@@ -98,3 +98,24 @@ git merge upstream/main  # or cherry-pick relevant commits
 
 > Add entries here when you discover a new pattern or fix a recurring mistake.
 > Format: `### <pattern name> (YYYY-MM-DD)`
+
+### Installer flatpak leaks to installed system (2026-06-01)
+
+The ISO's `install-flatpaks.sh` installs the bootc-installer as a system Flatpak into `/var/lib/flatpak/`. When fisherman runs `bootc install`, it copies all system flatpaks to the target — including the installer itself. The installed system then shows the installer as an available app.
+
+**Fix:** `bluefin-remove-installer.service` (a firstboot oneshot in `files/firstboot/`) removes `org.bootcinstaller.Installer` and `.Devel` if present, then prunes unused runtimes. Gated by a stamp file at `/var/lib/ublue-os/.installer-removed`.
+
+**Root cause is in `dakota-iso`** (`install-flatpaks.sh`), but the defensive fix lives in dakota because the installed image should never ship the installer regardless of how it got there.
+
+## Dakota vs Dakota-ISO boundary
+
+The installer is NOT built from source in this repo. The boundary:
+
+| What | Where |
+|------|-------|
+| OCI image (deployed to disk) | `projectbluefin/dakota` — this repo |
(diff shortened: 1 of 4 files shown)


dylanmtaylor:
Wouldn't it make more sense to fix this during the installation process so it's not there on first boot in the first place?

hanthor (approved):
LGTM. The `lab:pass` label confirms this has been validated on hardware.

**Service design is correct:**
- `ConditionPathExists=!/var/lib/ublue-os/.installer-removed` — stamp file gate prevents re-running on subsequent boots
- `After=local-fs.target` — correct ordering for flatpak + filesystem access
- `Type=oneshot` + `RemainAfterExit=yes` — right for a setup service
- Stamp file is always written at the end, regardless of whether an installer was found — means the cleanup only runs once total, which is the right behavior
- `|| true` on `flatpak uninstall --unused` is correct — runtime pruning failure shouldn't block boot
- BST wiring in `firstboot-date.bst` correctly installs the unit and symlinks it into `multi-user.target.wants/`

**docs/skills/installer.md** addition clearly documents the dakota-vs-dakota-iso boundary, which is the kind of boundary confusion that costs sessions.

PR #669, opened 2026-06-02
chore(agents): remove PR rate limit — gate is human now

castrojo:
Removes the hardcoded 4-PR open limit from AGENTS.md. PR volume is now gated by human maintainers, not an arbitrary agent-side cap.


hanthor (approved):
LGTM. Removing the hard 4-PR cap makes sense now that the queue system and human review are the gatekeeping mechanism. Clean diff, no functional changes to the build.

PR #743, opened 2026-06-08
feat(cc-panel): scaffold finupdate Updates panel for gnome-control-center

hanthor:
## Summary

Scaffolding to embed the [finupdate](https://github.com/hanthor/finupdate) **Software Updates** panel inside `gnome-control-center` so Dakota's Settings app gains a sidebar entry showing the booted image identity, a Check-for-updates button, changelog diff, and switch/pin actions.

The intent is to land this via the buildstream (no Containerfile overlay, per AGENTS.md) by patching `gnome-control-center` inside the `gnome-build-meta` junction.

## What's in this PR

- `elements/bluefin/finupdate.bst` — element that builds the `libfinupdate` Rust cdylib (`libfinupdate.so` + `finupdate.h` + `libfinupdate.pc`) from the finupdate repo. The panel's C code links against this via pkg-config.
- `files/finupdate-cc-panel/` — vendored panel sources copied verbatim from finupdate's `cc-panel/panels/updates/` (cc-updates-panel.{c,h,ui}, gresource xml, .desktop.in, meson.build, libfinupdate.pc.in).
- `docs/skills/finupdate-cc-panel.md` — integration design, link to the upstream cc-panel README, and the recommendation to patch via `patches/gnome-build-meta/` rather than overriding the whole cc element.

## What is _not_ in this PR (the gating work)

- [ ] **Pin libfinupdate to a release tag.** Currently `ref:` points at a `main` commit (`e5c393e`).
- [ ] **Cargo deps lock.** `finupdate.bst` has a commented-out `kind: cargo2` block — needs to be generated from finupdate's `Cargo.lock`.
- [ ] **gnome-control-center patch.** The panel only gets discovered if it's registered in `shell/cc-panel-loader.c`'s `default_panels` array and added to `panels/meson.build`. Recommended approach is a new patch under `patches/gnome-build-meta/` that drops the panel sources into `core/gnome-control-center.bst` and applies a small unified diff. See `docs/skills/finupdate-cc-panel.md` for the design tradeoffs.
- [ ] **fakecap manifest** updates for the new files under `core/gnome-control-center.bst`.

CI will fail at the bst build stage until at least the cargo2 deps and the cc-panel-loader patch are in place — opening non-draft per request so the failure surfaces concretely.

## Test plan

- [ ] `bst build elements/bluefin/finupdate.bst` succeeds once cargo2 deps are populated
- [ ] `bst build elements/bluefin/common.bst` (or whichever pulls cc) succeeds with the gbm patch applied
- [ ] In a built Dakota image: `gnome-control-center updates` opens the new panel
- [ ] Sidebar shows "Software Updates" entry with correct icon
- [ ] "Check" button reports up-to-date / update available against the registry

## Summary by CodeRabbit

* **New Features**
  * Added a "Software Updates" panel to GNOME Control Center, including UI, discovery/registration, and a packaged backend for showing system updates.

* **Documentation**
  * Added integration documentation describing embedding the updates panel and build/runtime expectations for the backend.


hanthor:
Good scaffolding. The approach (a) recommendation (patch_queue under patches/gnome-build-meta/) is the right call — it matches existing dakota patching patterns and keeps the panel changes co-located. The unchecked gating items are clearly enumerated and the 'intentionally opening non-draft so CI failure surfaces' is fine.

A few notes for the follow-up work:
- The 'ref: e5c393e' pointing at a main commit should be pinned to a tag before this can land in production — the TODO comment captures this correctly.
- The commented-out cargo2 block will need to be populated from finupdate's Cargo.lock before bst build succeeds.
- The gnome-control-center patch (approach a) is the critical path — the vendored panel sources land nothing without it.

Ready for another reviewer once the gating items are resolved.

hanthor:
Update: patches proven in a toolbox loop.

The `cc-panel-loader.c` + `panels/meson.build` edits are now in `patches/gnome-build-meta/finupdate-cc-panel-*.patch`. Generated by `just toolbox-test-cc-panel` in the finupdate repo, which builds an upstream `gnome-control-center` (gnome-49 against Fedora 43's stack) with the panel registered — launches cleanly, no broken-panel warnings. The same patches apply with no rejects against `gnome-50` (confirmed via `git apply --check`).

Also fixed the vendored panel meson.build (`desktopdir` → `control_center_desktopdir`, the actual cc variable name).

Note: dakota is on gnome-50 — the gnome-49 toolbox test was a pragmatic choice because Fedora 43's GTK/libadwaita stack matches gnome-49. The patches are version-agnostic for the bits they touch.

Still gating CI green:
- libfinupdate cargo2 deps in `finupdate.bst`
- override element wiring (pull patches + panel sources into `core/gnome-control-center.bst`) — see `docs/skills/finupdate-cc-panel.md`

castrojo:
Can we see some screenshots?

castrojo:
## Review — #743 feat(cc-panel): scaffold finupdate Updates panel for gnome-control-center

**Verdict: NOGO — convert to draft and resolve the critical issues below before re-requesting review.**

The core problem: `patch_queue` on `gnome-build-meta.bst` applies patches against **gnome-build-meta's own file tree** (BST YAML, `project.conf`, etc.) — exactly as the existing `disable-lorry-mirrors.patch` demonstrates. `shell/cc-panel-loader.c` and `panels/meson.build` live in the **gnome-control-center upstream source tree**, which is a completely separate git checkout fetched by `core/gnome-control-center.bst` inside gnome-build-meta. Those files don't exist in gnome-build-meta's tree at all — hence the CI error `No such file or directory`. This structural mismatch cascades into several other blocking issues.

---

### Critical (blocks merge)

**1. `patches/gnome-build-meta/` — patches target the wrong source tree.**
Both `finupdate-cc-panel-loader.patch` and `finupdate-cc-panel-meson.patch` diff against gnome-control-center C sources (`shell/cc-panel-loader.c`, `panels/meson.build`). The `patch_queue` source on `gnome-build-meta.bst` only sees gnome-build-meta's own files. The correct approach is to override `core/gnome-control-center.bst` *inside* gnome-build-meta by adding a `patch_queue` source there (via an entry in the `gnome-build-meta.bst` `overrides:` block), so the C-level patches are applied to the gnome-control-center source checkout. The design doc's "approach (a)" description is correct in intent, but the current patch routing doesn't implement it.

**2. `elements/gnome-build-meta.bst` — no override added for `core/gnome-control-center.bst`.**
Even if the patches were correctly routed, no `overrides:` entry was added to `elements/gnome-build-meta.bst`. The panel is unreachable from the build graph.

**3. `elements/bluefin/finupdate.bst` — CC panel is never compiled.**
`build-commands` only runs `cargo build --release --lib` for the Rust cdylib. The vendored C sources in `cc-panel-vendored/` have a complete `meson.build` but there are no `meson setup / meson compile / meson install` steps. `CcUpdatesPanel` is never built.

**4. `elements/bluefin/finupdate.bst` — element is not wired into the image.**
`finupdate.bst` doesn't appear in `elements/bluefin/common.bst` or `elements/bluefin/deps.bst`'s `depends:` list. BST will never schedule it and it will never land in the OCI image.

**5. `files/finupdate-cc-panel/libfinupdate.pc.in:8` + `finupdate.bst:73` — `@PREFIX@`/`@prefix@` case mismatch.**
The `.pc.in` template writes `prefix=@PREFIX@` (uppercase) but the `sed` in `install-commands` substitutes `-e "s|@prefix@|%{prefix}|g"` (lowercase). The regex never matches; the generated `.pc` will contain the literal `prefix=@PREFIX@`, breaking pkg-config resolution of `libfinupdate` at gnome-control-center build time. Also: `@libdir@`/`@includedir@` tokens referenced in the `sed` commands don't appear in the template, so those substitutions silently no-op.

---

### Important (should fix before merge)

- **`track: main` on upstream source (`finupdate.bst:31`)** — convention in this repo is to pin refs and not leave `track:` on a moving branch head. Use `track: gnome-50` to match the project cadence or remove `track` entirely with a comment on the bump procedure.
- **`docs/skills/README.md` not updated** — AGENTS.md requires new skill files to be registered in the routing table. `docs/skills/finupdate-cc-panel.md` was created but not added to the index.

---

### Minor

- `cc-updates-panel.h` uses `<shell/cc-panel.h>` (angle brackets) but `cc-updates-panel.c` uses `"shell/cc-panel.h"` (quoted). Be consistent — angle brackets are standard for public GNOME API headers.
- `finupdate_panel_widget_new()` failure path is silent. Add `g_warning()` so failures are visible in journald.
- Icon name `org.gnome.Settings-system-symbolic` is non-standard — verify it renders before landing.

---

### On the non-draft filing

AGENTS.md is unambiguous: *"No WIP PRs"* / *"Every PR must confirm `just lint` passed."* The author's note that CI will intentionally fail is understood, but the gate applies regardless. Please convert to draft until the three structural issues (patch routing, element wiring, cc-panel build) are resolved — then re-request review with a passing `validate` check.

castrojo (changes requested):
See review comment — 5 critical structural issues must be resolved before this can merge. Please convert to draft.

hanthor (commented):
Ambitious finupdate CC panel scaffold. Good docs and build integration. Check that gnome-control-center extension point is stable across GNOME versions.

hanthor (commented):
Found one issue before this can land:

**`@PREFIX@` vs `@prefix@` case mismatch** in `libfinupdate.pc.in` → `finupdate.bst`:

`libfinupdate.pc.in` line 7 uses uppercase:
```
prefix=@PREFIX@
```
But `finupdate.bst` install-commands substitute lowercase:
```
sed -e "s|@prefix@|%{prefix}|g"
```
These won't match — the generated `.pc` file ships with unsubstituted `@PREFIX@`.

Otherwise design looks solid. Docs are thorough, approach recommendation (option a) is well-reasoned, patches are clean and minimal. Fix the case mismatch and ship.

PR #746, opened 2026-06-08
fix(justfile): add seccomp=unconfined to just sbom podman invocations

castrojo:
## Problem

`just sbom` fails in CI with:
```
Error: crun: linkat `.cache/seccomp/...` to `container-id/seccomp.bpf`: Permission denied: OCI permission denied
```

## Root cause

crun 1.15+ caches compiled seccomp BPF programs via `linkat()`. Newer podman (installed via `update-podman: true` in setup-runner) no longer implies `seccomp=unconfined` from `--privileged`. When `buildstream-sbom` calls `bst show` inside the bst2 container, the generated source plugin triggers a nested crun invocation, and crun fails with EPERM trying to cache its seccomp BPF profile.

## Fix

Add `--security-opt seccomp=unconfined` to both `podman run` calls in the `sbom` recipe (prime step + generate step).

Fixes: https://github.com/projectbluefin/dakota/actions/runs/27141921221/job/80109170683

---

## Summary by CodeRabbit

* **Chores**
  * Updated container runtime configuration for internal build processes.

---

**Note:** This release contains infrastructure-level updates with no direct impact on user-facing functionality.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -944,6 +944,7 @@ sbom variant="default":
     echo "==> Priming BST generated source cache (${ELEMENT})..."
     podman run --rm \
         --privileged \
+        --security-opt seccomp=unconfined \
         --device /dev/fuse \
         --network=host \
         -v "{{justfile_directory()}}:/src:rw" \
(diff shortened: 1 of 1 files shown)


castrojo:
Superseded by #745 which was merged first with the same fix (different approach: drops --privileged entirely rather than adding seccomp=unconfined).

PR #750, opened 2026-06-08
Add keyring unlock service for Dakota

wiulinu:

---

## Testing

- [ ] `just validate` passes

## Checklist

````markdown
```verify
```
````

## Summary by CodeRabbit

* **New Features**
  * Added automatic unlocking of the GNOME login keyring during system startup, enabling seamless authentication for applications that depend on keyring access without requiring manual intervention.


hanthor (changes requested):
The intent is correct — auto-unlocking the keyring on boot is a real UX pain point. However the implementation has a significant technical issue:

**`gnome-keyring-daemon --daemon --launch-app` won't unlock an encrypted keyring.**
The keyring is encrypted with the user's login password. Running it as a system service (as root, before the display manager) has no access to the user session or the user's password — so the daemon starts but the 'login' keyring stays locked. Applications will still prompt for the password.

The standard approaches are:
1. **PAM integration** — `pam_gnome_keyring.so` in the PAM stack unlocks the keyring during GDM login using the session password. This is the correct upstream approach and should already work with GDM on Fedora. Check if it's being bypassed somewhere.
2. **Empty-password keyring** — Set an empty password so no unlock is needed. This is a deliberate security tradeoff and should be documented as such.
3. **User-session service** — If this must be a systemd unit, it needs to be a **user** service (`[Install] WantedBy=default.target` + `systemctl --user enable`), not a system service, so it runs in the user session context.

Additionally:
- The PR body is empty (template not filled in) — please describe what specific regression or issue this is fixing.
- No `User=` directive means the daemon runs as root, which is incorrect for gnome-keyring-daemon.
- The `--launch-app` flag is for D-Bus activation, not for session unlock.

Please investigate the PAM path first before adding a custom service.

PR #785, opened 2026-06-10
ci: add scheduled cache-warm workflow for BST remote cache

castrojo:
## Summary

Automation audit Phase 7: scheduled workflow to keep the BuildStream remote cache hot.

Addresses ND1 (cold-start timeout non-determinism) from the automation audit.

**Problem:** Dakota BST builds have a 360-min timeout. When cache misses after junction ref bumps or gbm upstream rebuilds, the full GNOME stack must rebuild from scratch — which can exceed 6h.

**Solution:** Schedule a cache-warm build Mon/Thu at 06:00 UTC to keep artifacts fresh before typical work-hour PR merge windows.

**Behavior:**
- Runs on schedule: Monday and Thursday 06:00 UTC
- Also available via `workflow_dispatch` for manual warm
- Uses `cancel-in-progress: false` — cache warm should complete even if a newer one starts
- Build failure is non-fatal (`|| true`) — this is warming, not gating

Assisted-by: Claude Sonnet 4.5 via pi

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,8 @@ on:
     branches: [main, next, testing]
   merge_group:
     branches: [main, next, testing]
+  push:
+    branches: [testing]
   workflow_dispatch:
     # WARNING: Do not manually dispatch immediately after auto/track-* ref bumps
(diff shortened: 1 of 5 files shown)


castrojo:
Superseded by clean branch from main.

PR #830, opened 2026-06-13
fix(ci): use @v1 managed tag for all projectbluefin/actions refs

castrojo:
## Problem

SHA-pinning internal org actions causes cascading reliability failures:

- A fix in `projectbluefin/actions` requires N separate Renovate bump PRs — one per consumer repo — before it takes effect anywhere
- During the lag, every consumer runs the broken version silently
- Tonight's automerge breakage (`--auto` bug, in since June 7) persisted for **6 days** because the SHA hadn't been bumped in `renovate-automerge.yml`
- `main` had a different SHA than `testing`, creating a split-brain where the two branches ran different workflow versions

We had **7 different SHAs** across 10 files, all pointing to different points in `projectbluefin/actions` history.

`AGENTS.md` already has the policy:
> `projectbluefin/` refs (`@v1`, `@main`) are intentional managed tags and are exempted.

It just wasn't being honored.

## Fix

1. **Converts all 17 `projectbluefin/actions` SHA refs to `@v1`** across 10 workflow files — fixes propagate to all consumers the moment they land in `actions`
2. **Adds `no-sha-pins-for-internal-actions` pre-commit hook** — blocks any future SHA pin to a `projectbluefin/` action at commit time

External action refs (`actions/checkout`, etc.) remain SHA-pinned as required by security policy.

## What `@v1` means

The `v1` tag is org-maintained and moves forward with every non-breaking fix. It currently points to `33a6406` (the `--auto` fix). `v1.1.0` / `v1.0.0` are point-in-time release tags that don't move.

## Summary by CodeRabbit

* **Chores**
  * Updated GitHub Actions workflow files across the CI/CD pipeline, including build, cache, release, publish, and sync workflows, to reference actions using semantic version tags instead of specific commit hashes.
  * Implemented new pre-commit validation hooks to enforce consistent action versioning and prevent floating action tag usage.


castrojo:
Updated with a second commit: adds `testing` to `build.yml`'s push trigger.

`publish.yml` already had the logic to publish `:testing` images on push to the testing branch — with a comment saying it should match bluefin/bluefin-lts behaviour — but `build.yml` only triggered on push to `[main, next]`. So `:testing` images were never actually updated after any Renovate merge landed.

Full pipeline after both commits land:
```
Renovate merge → push to testing
  → build.yml fires → build runs → OCI artifact produced
  → publish.yml fires (workflow_run) → :testing published ✓
  → promote-testing-to-main fires → promote PR updated ✓
  → human approves promote PR → merges to main
  → build.yml fires on push to main → publish → :latest/:stable ✓
  → sync-main-to-testing fires → testing kept current ✓
```

PR #886, opened 2026-06-16
fix(ci): pre-partition disk before bootc to avoid udev race on node creation

castrojo:
## Problem

PR #883 fixed `ioctl(BLKRRPART)` returning `EINVAL` by adding `-P` (PARTSCAN) to `losetup`. That got bootc further, but it now hits the next layer of the same race:

```
Creating root filesystem (xfs) on device /dev/loop0p3 (size=31.7 GB)
> mkfs.xfs -m uuid=... -L root /dev/loop0p3
error: Installing to disk: Creating rootfs: No such file or directory (os error 2)
```

**Root cause:** Device node creation from scratch is async. udevd processes the uevent `after` sfdisk returns. bootc's `mkfs.xfs` call immediately follows sfdisk and loses the race — `/dev/loop0p3` doesn't exist yet.

## Fix

Pre-create the GPT partition table with the exact same layout bootc will use, then `udevadm settle`. When nodes already exist, bootc's internal sfdisk performs an **in-place UPDATE** (synchronous) rather than a **CREATE FROM SCRATCH** (async udev event). mkfs finds the node immediately.

Partition type GUIDs and sector layout match bootc's observed x86-64 default.

## Checklist
- [x] `actionlint` passes
- [x] Builds on #883 (PARTSCAN foundation)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of boot image installation by pre-priming loopback partitions and adding a fail-fast check to verify the root filesystem is created, with clearer failure diagnostics.

* **Documentation**
  * Expanded troubleshooting guidance for loopback-based install issues, including race-condition mitigation steps and validation checks.
  * Clarified CI promotion workflow quality-gate behavior and action-tag pinning policy.

* **Chores**
  * Tightened repository checks for GitHub Action `uses:` references to better enforce tag and SHA pinning rules.

### DIFF
--- a/.github/workflows/promote-testing-to-main.yml
+++ b/.github/workflows/promote-testing-to-main.yml
@@ -37,4 +37,8 @@ jobs:
       cosign_identity_regexp: >-
         ^https://github\.com/projectbluefin/(dakota|actions)/\.github/workflows/
       run_e2e: false
+      # run_e2e is intentionally false for dakota. The e2e quality gate lives
+      # at the weekly-testing-promotion.yml level with a GitHub 'production'
+      # Environment requiring 2 explicit human approvals — not at the PR gate
+      # level. The promotion PR here gets cosign verification only.
     secrets: inherit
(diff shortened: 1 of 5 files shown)


castrojo:
Superseded by #889 (clean rebase onto main after #883 merge)

Issue #1163, opened 2026-07-15
Official GNOME system-monitor extension silently fails to load because of missing dependency


Raindrac:
### Before submitting

- [x] I searched existing issues and this is not a duplicate
- [ ] I ran ujust report (or am about to — paste the URL below)

### What happened?

The official GNOME system-monitor shell extension silently fails to load because of a missing dependency. Some light research indicates that it is likely [libgtop](https://github.com/GNOME/libgtop).

From the extension's `.js` file:
```js
import GTop from 'gi://GTop';
```

Normally I wouldn't create an issue for a missing dependency from external software, but in this case it's an official GNOME extension so I figure it might as well be addressed. I wasn't able to get it working via containerizing the lib.

### Steps to reproduce

1. Open Extensions Manager
2. Search for "System Monitor" by fmuellner, which is the official extension (https://gitlab.gnome.org/GNOME/gnome-shell-extensions)
3. Observe how it never loads

### Affected area

GNOME / desktop

ahmedadan:
Hi @Raindrac,

I am unable to reproduce on Dakota Testing with GNOME Shell 50.2 and System Monitor 50.2. The extension enters ACTIVE state, reloads without errors, and its GTop CPU/memory/swap/network calls work. The image contains both the GTop typelib and libgtop runtime, pulled transitively through GNOME Control Center.

Could you provide a `ujust report`, and the output of:

```bash
gnome-extensions info system-monitor@gnome-shell-extensions.gcampax.github.com
```

and any relevant GNOME Shell errors from:

```bash
journalctl -b _COMM=gnome-shell --no-pager | grep -iE 'system-monitor|GTop|JS ERROR|extension.*(fail|error)'
```

ahmedadan:
While digging into this I found the typelib only reaches the image transitively, gnome-control-center happens to depend on libgtop and nothing in our own manifest asks for it. I've made it an explicit dependency so an upstream dep shuffle can't silently drop it. I'm still curious about your case though, the info from my earlier comment would help pin down whether you were on an image from a bad window.

Raindrac:
Here is the `ujust report`: https://gist.github.com/Raindrac/93b52e6058f98d57f91be384a4415997

The output of `gnome-extensions info system-monitor@gnome-shell-extensions.gcampax.github.com`:
```sh
>  gnome-extensions info system-monitor@gnome-shell-extensions.gcampax.github.com
system-monitor@gnome-shell-extensions.gcampax.github.com
[...]

PR #1171, opened 2026-07-20
fix(bluefin): declare libgtop explicitly for GTop-2.0 typelib

ahmedadan:
## Summary

Hardens against the system-monitor Shell extension losing its `GTop-2.0.typelib` (#1163). The typelib currently reaches the image only transitively: `gnomeos-deps/deps.bst -> meta-gnome-core-shell -> gnome-control-center -> core-deps/libgtop`. Nothing in Dakota's own element graph declares it, so any upstream refactor that drops libgtop from gnome-control-center's dependencies (or drops gnome-control-center from the core-shell meta) would silently remove the typelib and break the extension.

Adds `gnome-build-meta.bst:core-deps/libgtop.bst` explicitly to `bluefin/deps.bst` with a comment stating the constraint. The full transitive chain is documented in the commit body.

Verified: `bst show` resolves the junction element and the full default-variant image build passes. The typelib already ships today via the transitive chain; this change only pins it, so it cannot regress anything. This is preventive hardening rather than a repair, the issue itself is pending reporter info (likely a stale image), hence Related rather than Fixes.

Related: projectbluefin/dakota#1163

### DIFF
--- a/elements/bluefin/deps.bst
+++ b/elements/bluefin/deps.bst
@@ -45,6 +45,10 @@ depends:
   # Include things missing from the gnomeos base image
   - gnome-build-meta.bst:core-deps/fwupd.bst
 
+  # system-monitor shell extension needs GTop-2.0.typelib; only reaches the image
+  # transitively via gnome-control-center otherwise (projectbluefin/dakota#1163)
+  - gnome-build-meta.bst:core-deps/libgtop.bst
+
   - freedesktop-sdk.bst:components/buildstream2.bst
   - freedesktop-sdk.bst:components/containers-common.bst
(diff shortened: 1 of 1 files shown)


hanthor (approved):
Approved. Low-risk preventive hardening: pins core-deps/libgtop.bst explicitly in bluefin/deps.bst so GTop-2.0.typelib no longer reaches the image only transitively via gnome-control-center. The typelib already ships today, so this only removes a latent regression path and cannot break anything. Comment documents the constraint. Mergeable; validate + CodeRabbit green. Treating ghost-lab as non-gating (pending across all open PRs).

castrojo:

PR #1172, opened 2026-07-20
fix(motd): restore MOTD display, drop orphaned legacy motd

### LINKED ISSUE #930: motd env missing
### Before submitting

- [x] I ran ujust report (or am about to — paste the URL below)

### ujust report gist URL

https://gist.github.com/bketelsen/d10618afc3e4b4eb82bd791b2b198f59

### What happened?

any new terminal gets this error: 
```/usr/bin/ublue-motd: line 12: /usr/share/ublue-os/motd/env.sh: No such file or directory```

### Steps to reproduce

open a terminal

### Affected area

GNOME / desktop

### Extra context

_No response_

ahmedadan:
## Summary

Fixes MOTD not displaying at all on current images, the follow-on regression found while verifying the original #930 env fix. Tracing it back: [c9bac80](https://github.com/projectbluefin/dakota/commit/c9bac802ebb41e21641e31c9417edbe27f512b01) moved MOTD from motd.bst to umotd, and one day later [53480cd](https://github.com/projectbluefin/dakota/commit/53480cd0b10fd1d7144d3bc5b022c4dbeca94f74) added a build rule stripping common's `umotd.sh` profile hook, citing double-display with motd.bst as the reason. motd.bst was already gone by then, so the rule deletes the only hook that invokes umotd and nothing ever runs it.

Two commits:

1. **Remove the stale `rm -f` in `common.bst`** so common's `umotd.sh` profile hook survives into `/etc/profile.d/`. Verified at the pinned common ref that the hook invokes `umotd` unconditionally (no `BLUEFIN_SHELL_ENABLE_MOTD` gate in the hook or the binary), so the fix is live, not inert. The legitimate rechunker-group-fix strip block next to it is untouched.
2. **Delete the orphaned `motd.bst` and `files/motd/`**, verified unreferenced by a repo-wide sweep of the element graph, Justfile, scripts, and workflows.

With motd.bst gone there is nothing left to double up with, so restoring the hook cannot reintroduce the double-MOTD the stale rule was originally written against.

Verified: full image builds clean. Boot check for exactly-one-MOTD lands with the published image, tracked in the issue.

Fixes: projectbluefin/dakota#930

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed leftover Bluefin migration services, executables, and system presets from installed images.
  * Preserved the standard system profile hook instead of explicitly removing it.

* **Changes**
  * Removed the custom Dakota welcome message and its supporting content.
  * The image no longer installs or displays the previous Bluefin MOTD template, tips, or related environment information.


hanthor (approved):
Approved. Correctly diagnoses the regression: the `rm -f .../umotd.sh` in common.bst was added citing double-display with motd.bst, but motd.bst was already gone, so the rule deleted the only hook that invokes umotd. Removing it restores the profile hook. Verified the deleted motd.bst is not referenced as a build dependency: deps.bst depends on `bluefin/umotd.bst` (a different element), and the only motd.bst reference was the common.bst comment this PR removes. The adjacent legitimate rechunker-group-fix strip block is untouched. Mergeable; validate + CodeRabbit green. Treating ghost-lab as non-gating (pending across all open PRs).

castrojo:

PR #1173, opened 2026-07-20
fix(bluefin): ship VTE typelibs for GTK3 GI consumers

### LINKED ISSUE #1022: Vte-2.91.typelib missing
### Before submitting

### ujust report gist URL

_No response_

### What happened?

Image
Trying to add this beautiful ddterm extension and got this error.

### Steps to reproduce

Install ddterm via Externsion Manager.

### Affected area

GNOME / desktop

### Extra context

I also install Freedesktop SDK, thought it contains all that required libs, rebooted, and stacked at this error again.
// And that dosu thing is really helpful but wants pro subscription:
You've reached the limit of 20 interactions per month. Your usage will reset on 2026-07-01. If you'd like to use Dosu before then, upgrade to a pro account.

ahmedadan:
## Summary

Fixes the ddterm GNOME Shell extension failing to load because the image ships no VTE typelibs at all (#1022). Dakota's `meta-gnome-core-apps` override drops gnome-console, gnome-builder, and foundry, the only things in the upstream stack that pull vte in, and upstream's `core-deps/vte.bst` builds GTK4-only anyway while ddterm binds the GTK3 ABI.

New `bluefin/vte.bst` rebuilds the identical vte 0.84.0 source with `-Dgtk3=true -Dgtk4=true`, staging both ABIs:

1. **`Vte-2.91.typelib` + `libvte-2.91.so.0`** (GTK3, what ddterm needs)
2. **`Vte-3.91.typelib` + `libvte-2.91-gtk4.so.0`** (GTK4, parity with upstream)

The source ref is pinned in lockstep with the gnome-build-meta junction's vte and commented as a bump-together constraint, so we never ship two VTE ABI generations. Split rules push the demo apps, both sets of demo `.desktop` registrations (including the `xdg-terminals/` entries that would otherwise dangle), the unversioned linker symlinks, and the `.gir` XML to devel, so only the typelibs and versioned libraries reach the composed image. docs/vapi/glade are disabled to avoid pulling gi-docgen/vala for artifacts the compose discards.

Verified: element and full default-variant image builds pass, and the composed `bluefin-runtime` layer contains exactly the two typelibs plus versioned libs with no demo/desktop/gir leakage.

Fixes: projectbluefin/dakota#1022

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored VTE runtime libraries and introspection data needed by applications and desktop extensions.
  * Improved compatibility for terminal-related functionality, including GTK3 and GTK4 environments.
  * Ensured required runtime components are included without unnecessarily packaging development files.

* **Documentation**
  * Added guidance for preserving transitive libraries and introspection files when customizing application contents.
  * Documented recommended patterns for supporting GTK3 introspection and runtime packaging.

### DIFF
--- a/docs/skills/add-package.md
+++ b/docs/skills/add-package.md
@@ -158,3 +158,18 @@ variables:
 ```
 
 External dep: `freedesktop-sdk.bst:components/sndfile.bst`. No external LV2 headers needed — bundled.
+
+### Dropping upstream apps silently drops their transitive libs/typelibs (2026-07-19)
+
+`core/meta-gnome-core-apps.bst` is overridden to a short allow-list, so any library
+pulled *only* by a dropped app never enters the image. This is invisible until a GI
+consumer fails to bind it. Example: dakota#1022 — gnome-console/builder/foundry were
+the only pullers of gnome-build-meta's `core-deps/vte.bst`, so the image shipped **zero**
+VTE typelibs and the user-installed ddterm extension (GTK3 ABI, `Vte-2.91`) could not load.
+
+Fix pattern: add a dedicated `bluefin/<lib>.bst` that rebuilds the *same source/ref as
+the junction element* (keep them in lockstep — bump together) rather than overriding the
+junction. If a Shell extension needs the GTK3 introspection ABI, build with `-Dgtk3=true`
+(depends on `gnome-build-meta.bst:sdk/gtk+-3.bst`); upstream's VTE is GTK4-only. Push demo
+binaries and unversioned `.so` linker symlinks to the `devel` split so `bluefin-runtime`'s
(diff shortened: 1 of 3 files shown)


hanthor (approved):
Approved. Correctly restores the VTE typelibs dropped when gnome-console/builder/foundry were removed from the app allow-list. Rebuilds the identical 0.84.0 source (lockstep constraint documented) with -Dgtk3=true -Dgtk4=true so ddterm's GTK3 ABI (Vte-2.91) is present, and the split-rules push demo bins, .desktop/xdg-terminals registrations, unversioned linker symlinks, and .gir XML to devel so only typelibs + versioned .so.N reach the runtime compose. Mergeable; validate + CodeRabbit green. Treating ghost-lab as non-gating (pending across all open PRs).

castrojo:

PR #1174, opened 2026-07-20
fix(bluefin): ship alsa-utils to restore mixer state on boot

### LINKED ISSUE #1027: Dakota boots with broken sound
### Before submitting

### ujust report gist URL

https://github.com/projectbluefin/dakota/issues/new?template=bug-report.yml

### What happened?

Dakota boots with broken sound. The system does not remember the settings after reboot.

### Steps to reproduce

1. Reboot the system
2. Try to play sound

### Affected area

GNOME / desktop

### Extra context

_No response_

ahmedadan:
## Summary

Fixes "boots with broken sound" (#1027). freedesktop-sdk ships alsa-lib but no alsa-utils, so the image has no `alsactl`, no restore units, no udev restore rule, and no ALSA init database. Kernel mixer state therefore resets to hardware power-on defaults on every boot, and on codecs whose defaults are muted or mis-routed (auto-mute, amp-enable, IEC958, output selection) that presents exactly as broken sound at boot.

New `bluefin/alsa-utils.bst` ships the mixer-state machinery and nothing else:

1. **`alsactl` plus its systemd units**, enabled via the upstream static `sound.target.wants/` symlinks that `make install` creates. The units carry no `[Install]` section, so a systemd preset cannot enable them; the static wants mechanism is preset-all-safe and already the repo convention for bindmounts.
2. **`90-alsa-restore.rules`** restores per-card state on hotplug, and the **`/usr/share/alsa/init` database** initializes default-muted hardware sanely on first boot before any saved state exists.
3. **Merged-usr handling**: `alsactl` is a `sbin_PROGRAM` and this image has `/usr/sbin -> bin`, so the element sets `--sbindir=%{bindir}` to avoid colliding with the symlink at compose time while keeping every `@sbindir@` token in units and rules resolving correctly.
4. **Version lockstep**: alsa-utils is pinned to 1.2.14 to match fdsdk's alsa-lib 1.2.14 (1.2.16 fails to compile against it, `snd_lib_log_*` only exists in alsa-lib 1.2.16). Commented as a bump-together constraint.

First-boot behavior is safe by upstream design: `ExecStart=-alsactl restore` ignores a missing state file, and the save half (`ExecStop` store on shutdown) creates `/var/lib/alsa/asound.state` for subsequent restores. The rest of the suite (amixer, aplay, seq/midi tools, docs, translations) is pruned to keep the image lean.

Verified: element and full default-variant image builds pass, and the artifact contains exactly alsactl, the two units with their wants symlinks, the udev rule, and the init database. Needs verification on affected hardware once published, tracked in the issue.

Fixes: projectbluefin/dakota#1027

### DIFF
--- a/elements/bluefin/alsa-utils.bst
+++ b/elements/bluefin/alsa-utils.bst
@@ -1,0 +1,85 @@
+kind: autotools
+description: |
+  ALSA mixer state save/restore.
+
+  freedesktop-sdk ships alsa-lib (libasound) but no alsa-utils, so the image
+  has no alsactl, no alsa-restore/alsa-state systemd units and no
+  90-alsa-restore.rules. Nothing restores the kernel mixer state across
+  reboots, and the /usr/share/alsa/init initialisation database is absent.
+  On hardware whose power-on mixer defaults are muted or mis-routed
+  (auto-mute, amp-enable, IEC958, output selection) this surfaces as
+  "boots with broken sound" (projectbluefin/dakota#1027).
+
+  We build alsa-utils and keep only the mixer-state machinery: the alsactl
+  binary, its systemd units (statically enabled via sound.target.wants/ by
+  `make install`), the udev rule that restores per-card state on hotplug, and
+  the init database.
+
+sources:
+- kind: tar
+  # Canonical upstream release tarball (alsa-project.org). ref is the sha256
+  # of alsa-utils-1.2.14.tar.bz2, verified by fetching the tarball and running
+  # sha256sum.
+  #
+  # LOCKSTEP CONSTRAINT: alsa-utils and alsa-lib release together and the utils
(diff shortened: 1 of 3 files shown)


hanthor (approved):
Approved. Coherent fix for the missing mixer-state machinery: ships only alsactl plus its static sound.target.wants/ units (correct, since the units have no [Install] section and presets can't enable them), the udev restore rule, and the init database. The merged-usr handling (`--sbindir=%{bindir}` so @sbindir@ tokens resolve without colliding with /usr/sbin -> bin) and the 1.2.14 lockstep with fdsdk's alsa-lib are both sound and documented. Prune loop keeps the image lean. Mergeable; validate + CodeRabbit green. Treating ghost-lab as non-gating (pending across all open PRs).

castrojo:

PR #1175, opened 2026-07-20
feat(bluefin): enable fingerprint auth for sudo and polkit

### LINKED ISSUE #953: Fingerprint authentication not working for sudo or GNOME polkit prompts
### Before submitting

- [x] I ran ujust report (or am about to — paste the URL below)

### ujust report gist URL

https://gist.github.com/hesaratha/a4d37bba7b53b2e24a3399cca9fd9440

### What happened?

Fingerprint authentication is not available for privilege escalation. When prompted for auth, both by sudo in the terminal and in GNOME Settings the fingerprint scanner is not offered as an option.

fprintd is running when checked via systemctl.
Fingerprint authentication seems to work fine for unlocking from lock screen so I suspect its a PAM issue and not a hardware one.
Journalctl shows pam_unix failures for polkit-1 which might be relevant, or maybe PAM just isn't configured correctly.

Hardware: ThinkPad T14 Gen 3
Fingerprint scanner: Shenzhen Goodix Technology Co.,Ltd. Goodix USB2.0 MISC

### Steps to reproduce

1. Open Ptyxis and run: sudo ls
2. Observe: place finger on reader to no avail
3. Open GNOME Settings -> Users -> Unlock...
4. Observe: Again, no prompt to use the fingerprint, nor does placing finger on reader do anything
Image

### Affected area

other

### Extra context

Fingerprint authentication used to work fine on the Silverblue variant of Bluefin.
Authenticating sudo doesn't work on Ghostty either, but I wasn't sure if it didn't support it in the first place which is why I tried Ptyxis.

ahmedadan:
## Summary

Fixes fingerprint auth never being offered by sudo or polkit despite working enrollment and GDM login (#953). Enrollment and GDM work because `fingerprint-auth`/`gdm-fingerprint` reference `pam_fprintd` directly, but sudo and polkit reach auth through `auth include system-auth`, and fdsdk's `system-auth` has no fingerprint line. There is no authselect on this image, so nothing ever added one.

New `bluefin/fprintd-system-auth.bst` ships a full `/etc/pam.d/system-auth` override (PAM has no drop-in mechanism for the aggregate stack) with one added line: `auth sufficient pam_fprintd.so` above `pam_unix`.

1. **Ordering contract documented in the element**: fingerprint is offered first, and any failure (no match, no reader, nothing enrolled) falls through to password, so it can never lock anyone out.
2. **sudo-rs compatible**: the in-flight sudo-rs (built without the `pam-login` feature) opens the `sudo` PAM service for both `sudo` and `sudo -i`, and libpam expands the `include`, so the fix covers classic sudo, sudo-rs, and polkit (`/usr/lib/pam.d/polkit-1` also includes system-auth) through one file. The latent `pam-login`/`sudo-i` trap is documented in the commit and in `docs/skills/bst-overrides.md`.
3. **No reader, no delay**: pam_fprintd returns `PAM_AUTHINFO_UNAVAIL` immediately when fprintd reports no devices, so reader-less machines see no added latency.
4. **Zero drift at introduction**: the override is a verified one-line diff against the stock file, and the element documents the re-sync obligation at every fdsdk bump plus the exit condition (drop when upstream ships a fingerprint-enabled system-auth or the image gains authselect).

Verified: full default-variant image build passes, and the composed layer's `/etc/pam.d/system-auth` contains the fingerprint line, proving the override wins the overlap (element depends on `components/linux-pam.bst` to stage after the stock file, with the expected overlap whitelisted). Needs verification on fingerprint-reader hardware once published, tracked in the issue.

Fixes: projectbluefin/dakota#953

### DIFF
--- a/docs/skills/bst-overrides.md
+++ b/docs/skills/bst-overrides.md
@@ -290,3 +290,44 @@ Applying any local patch queue (`patch_queue` source) to a junction (like `gnome
 
 - **Consequence:** Carrying local patches (like `disable-lorry-mirrors.patch`) on a junction silently forces local compiles for massive components (like WebKit) by preventing cache reuse against the official public upstream cache (`gbm.gnome.org:11003`).
 - **Fix:** Keep junctions 100% clean of downstream patch queues. If a patch is required, submit it upstream first or bump the junction ref. Removing the patch queue on `gnome-build-meta` immediately restored 1053 out of 1090 cached elements (96% cache hit rate).
+
+### Overriding a single file a junction ships into /etc (overlap-whitelist pattern) (2026-07-19)
+
+To replace one file that an upstream junction component installs (not patch the
+junction, not shadow the whole element), ship a Dakota `bluefin/*.bst` element
+that installs the replacement to the same path, and use `overlap-whitelist` to
+win the compose overlap. This is the same mechanism `bluefin/sudo-rs.bst` uses
+to take over `/usr/bin/sudo` from fdsdk's `components/sudo.bst`.
+
+```yaml
+public:
+  bst:
+    overlap-whitelist:
+    - /etc/pam.d/system-auth   # literal path, not a %{var}
+depends:
+- freedesktop-sdk.bst:components/linux-pam.bst   # forces staging AFTER the stock file
+```
+
+Two facts that make this safe and deterministic:
+
+- **Staging order = overlap winner.** Whitelisting only silences the warning; the
+  element staged *later* wins. A runtime `depends:` on the component that ships
+  the original guarantees the override stages after it. (`sudo-rs.bst` gets away
+  without the explicit dep, but declaring it removes the ambiguity.)
+- **/etc vs /usr/etc matters.** fdsdk `sysconfdir=/etc`, so config like
+  `linux-pam`'s pam.d lands in real `/etc`, and the compose includes `/etc`.
+  `oci/bluefin.bst` only merges `/usr/etc -> /etc` (`cp -a`, `/usr/etc` wins) —
+  it never touches files already in `/etc`. So an `/etc` override is NOT
+  clobbered by that merge. If you instead need to override something a component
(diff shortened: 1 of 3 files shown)


hanthor (commented):
Neutral comment, not an approval — flagging for maintainer sign-off because this is a security-sensitive auth change.

The construction is correct and fail-open: `auth sufficient pam_fprintd.so` placed above pam_unix means any fprintd outcome (no match, no reader, nothing enrolled) falls through to the password, so it cannot lock anyone out. This matches the standard authselect `with-fingerprint` pattern, and pam_fprintd returns PAM_AUTHINFO_UNAVAIL immediately on reader-less hardware so there's no added latency. The /etc (not /usr/etc) placement correctly survives the OCI merge, and the overlap-whitelist + explicit linux-pam dependency make the override win deterministically.

Two things a human should own before merge:

1. Policy: accepting that a fingerprint may now authorize `sudo` and polkit (not just GDM login) is a security-posture decision, not just a code correctness one.
2. Fidelity: this is a hand-copied full-file override of fdsdk's stock `system-auth`, so every non-fingerprint line must stay byte-faithful to upstream — I can't diff it against the pinned fdsdk ref from here. Please confirm the stock stack matches, and note the re-sync obligation at every fdsdk bump the element already documents.

Also needs the on-hardware verification the PR calls out (reader-present and reader-less). CI: mergeable, validate + CodeRabbit green; ghost-lab treated as non-gating (pending across all open PRs).

hanthor (approved):
Source-level verification resolves my earlier fidelity concern: the pinned freedesktop-sdk 25.08.13 system-auth file matches this override byte-for-byte apart from the intended pam_fprintd line, so the full-file copy does not drop or alter another PAM control. The sufficient ordering preserves password fallback. Approving the implementation; the reader-present and reader-less hardware verification should remain tracked as stated.

castrojo:

PR #1176, opened 2026-07-20
feat(bluefin): migrate legacy /var/home passwd home fields to /home

### LINKED ISSUE #1170: Installs older than 2026-06-29 have stale /var/home passwd entries: snapd broken, path-identity bugs
### What happened?

#1117 turned /home and /root into real directories bind-mounted from /var/home and /var/roothome (snapd requires homes literally under /home) and switched user creation to useradd-default, so new users get `/home/<user>` in /etc/passwd. But /etc persists across image upgrades, so anyone who installed between ~2026-04-17 and 2026-06-29 still carries the old-layout home fields no matter how current their image is.

For that install cohort:
- snapd is broken because snap-confine refuses home directories outside /home, which defeats the feature #1117 restructured the filesystem to enable. These reports won't reproduce on any fresh install.
- Path-identity bugs are now possible. The old symlink layout canonicalized both spellings to the same path, but a bind mount does not, so `$HOME` from passwd and a derived `/home/<user>` path are two different absolute paths to the same content, and any path-equality logic between them silently disagrees.
- Machines can end up with mixed cohorts. A pre-June-29 install that adds a user today gets one /var/home user and one /home user side by side, and bugs become per-user.
- The divergence compounds over time. Every login bakes the literal `$HOME` string into more user state (shell rc, .desktop files, user units), so waiting makes the migration more expensive.

### Steps to reproduce

1. Take any Dakota install created between ~2026-04-17 and 2026-06-29 (or recreate one: on a current image, set a user's passwd home field to `/var/home/<user>` the way useradd-ostree did).
2. Upgrade to the current testing image.
3. Run `grep -E ':/var/(home|roothome)' /etc/passwd`. Any output means the stale entries are present.
4. Try any strictly confined snap as that user; snap-confine refuses to run ("home directories outside of /home are not currently supported").

### Affected area

build system / BST

### Extra context

Proposed fix: ship a one-shot idempotent migration in the image. A oneshot unit ordered `Before=systemd-user-sessions.service` (and accounts-daemon / display-manager) rewrites `/var/home/<user>` home fields to `/home/<user>` and root's `/var/roothome` to `/root` via `usermod -d`, which takes the canonical passwd lock and writes atomically. No data moves, the bind mount means both paths already reference the same directory, only the passwd string changes. Fast no-op on every boot once entries are clean, self-healing if a stale entry is ever reintroduced.

Implementation with standalone test transcript is on the `feat/var-home-passwd-migration` branch, PR to follow.

Out of scope, worth release-noting: dotfiles and configs that hardcode `/var/home/...` can't be auto-fixed by this migration, and that includes home data migrated from Fedora Atomic systems where /var/home is the canonical spelling. A doctor check that greps common config locations could warn about these.

Related: #1117

ahmedadan:
## Summary

Fixes stale `/var/home/<user>` passwd entries on installs created before 2026-06-29 (#1170). PR #1117 turned /home and /root into real directories bind-mounted from /var/home and /var/roothome (snapd requires homes literally under /home) and switched user creation to useradd-default. But /etc persists across image upgrades, so the pre-#1117 install cohort still carries old-layout home fields: snapd refuses to run for them, and because a bind mount (unlike the old symlink) does not canonicalize the two spellings together, `$HOME` from passwd and derived `/home/<user>` paths silently disagree in any path-equality logic.

New `bluefin/migrate-var-home-passwd.bst` ships a one-shot, idempotent boot-time migration:

1. **Ordering is the safety mechanism**: the oneshot unit runs `Before=systemd-user-sessions.service` (the barrier every display manager and getty orders after), plus `Before=accounts-daemon.service` and the display manager, so no session exists and nothing else has passwd open when it writes.
2. **The write is delegated to `usermod -d`**, which takes the canonical passwd lock (lckpwdf on `/etc/.pwd.lock`, the same lock vipw/passwd/accounts-daemon use), writes atomically, and maintains the `/etc/passwd-` backup. No hand-rolled locking: flock(2) does not interoperate with lckpwdf's fcntl lock, so reusing the canonical tool is strictly safer.
3. **No data moves**: the bind mount means both spellings already reference the same directory, only the passwd string changes (`/var/home/<user>` to `/home/<user>`, root's `/var/roothome` to `/root`).
4. **Idempotent and self-healing**: a fast grep exits 0 before taking any lock when nothing matches, and because there is deliberately no sentinel file, a stale entry reintroduced by a later /etc edit gets fixed on the next boot. One journal line per migrated user.

The script supports a `MIGRATE_PREFIX` override (passed straight to `usermod --prefix`) for standalone testing, and was tested that way against a copied passwd tree: multiple users, root, already-migrated entries, non-home system accounts, and custom home basenames all handled correctly, with a clean idempotent second run.

Out of scope, release-note material: user dotfiles that hardcode `/var/home/...` cannot be auto-fixed, including home data migrated from Fedora Atomic systems where /var/home is the canonical spelling.

Verified: element builds clean and the full default-variant image build passes. Boot verification on an affected install is tracked in the issue.

Fixes: projectbluefin/dakota#1170
Related: projectbluefin/dakota#1117

### DIFF
--- a/elements/bluefin/deps.bst
+++ b/elements/bluefin/deps.bst
@@ -64,5 +64,6 @@ depends:
   - gnome-build-meta.bst:gnomeos-deps/wl-clipboard.bst
 
   - bluefin/bindmounts.bst
+  - bluefin/migrate-var-home-passwd.bst
 
   - gnome-build-meta.bst:gnomeos-deps/snapd.bst
--- a/elements/bluefin/migrate-var-home-passwd.bst
+++ b/elements/bluefin/migrate-var-home-passwd.bst
@@ -1,0 +1,33 @@
+kind: manual
+description: |
+  One-shot, idempotent boot-time migration of legacy /var/home/<user> (and
+  /var/roothome) passwd home fields to the post-#1117 bind-mount layout.
+  Installs from before 2026-06-29 carry the old fields, which breaks snapd
+  and makes path-identity comparisons disagree; see
+  projectbluefin/dakota#1170.
+
+  Ships the migration script, its oneshot unit (ordered before
+  systemd-user-sessions), and a preset to enable it.
+
+sources:
+- kind: local
+  path: files/migrate-var-home-passwd
+
+build-depends:
+- freedesktop-sdk.bst:public-stacks/runtime-minimal.bst
+
+variables:
+  strip-binaries: ""
+
+config:
+  install-commands:
+  - |
+    install -Dm755 bluefin-migrate-var-home-passwd \
+            "%{install-root}/usr/libexec/bluefin-migrate-var-home-passwd"
+  - |
+    install -Dm644 bluefin-migrate-var-home-passwd.service \
+            "%{install-root}%{indep-libdir}/systemd/system/bluefin-migrate-var-home-passwd.service"
+  - |
+    install -Dm644 /dev/stdin "%{install-root}%{indep-libdir}/systemd/system-preset/80-bluefin-migrate-var-home-passwd.preset" <<'PRESET'
+    enable bluefin-migrate-var-home-passwd.service
+    PRESET
--- a/files/migrate-var-home-passwd/bluefin-migrate-var-home-passwd
+++ b/files/migrate-var-home-passwd/bluefin-migrate-var-home-passwd
@@ -1,0 +1,69 @@
+#!/usr/bin/bash
+# One-shot migration of legacy home fields in /etc/passwd: /var/home/<user>
+# -> /home/<user>, /var/roothome -> /root. Installs from before PR #1117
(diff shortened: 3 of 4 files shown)


hanthor (approved):
Approved. The migration is well-constructed: `usermod -d` (no `-m`) delegates locking/atomic-replace to canonical lckpwdf, the grep fast-path exits before locking when nothing matches, and reading the passwd FD open across the loop keeps every decision against a consistent pre-migration snapshot while usermod renames new inodes in. Idempotent, no sentinel (self-healing), and the unit ordering (Before=systemd-user-sessions/accounts-daemon/display-manager) ensures no session has passwd open during the write. Mergeable; validate + CodeRabbit green. Treating ghost-lab as non-gating (pending across all open PRs).

castrojo:

PR #1196, opened 2026-07-24
chore(workflow): pin bonedigger lifecycle to clanker-queue rollout

castrojo:
Pins dakota's bonedigger caller to the new bonedigger rollout SHA so it can consume the clanker-queue migration workflow.

## Summary by CodeRabbit

* **Chores**
  * Updated the automated lifecycle workflow to use a newer workflow revision.
  * No changes to workflow triggers, permissions, or job configuration.

### DIFF
--- a/.github/workflows/bonedigger.yml
+++ b/.github/workflows/bonedigger.yml
@@ -17,7 +17,7 @@ permissions:
 
 jobs:
   lifecycle:
-    uses: projectbluefin/bonedigger/.github/workflows/lifecycle.yml@d53076732c23203efd7856eaee7c502742897603 # v1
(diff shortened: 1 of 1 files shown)


hanthor (changes requested):
This one-line rollout change targets main from a branch based on the active testing line, so the PR currently contains a large set of unrelated workflow, element, patch, and documentation changes and is conflicting. Dakota content PRs must target testing. Please retarget or recreate it from testing with only the bonedigger pin, then refresh that pin after the blocking changes on projectbluefin/bonedigger#33 are resolved.

castrojo:
Merged into testing in commit a2feac9.

PR #1198, opened 2026-07-25
fix(build): set GOROOT for remote Go builds

castrojo:
## What
Set the installed freedesktop-sdk Go root explicitly for Dakota elements that compile Go sources. BuildBarn remote actions currently find the trimmed `go` binary but fail because `GOROOT` is unset.

This covers both the distrobox build that currently blocks the merged testing build and the NVIDIA container toolkit build.

## Verification
- `git diff --check` passed
- `just bst show oci/bluefin.bst` passed
- The failed lab RE build showed `go: cannot find GOROOT directory` in `bluefin/distrobox.bst`
- Lab RE rebuild and boot/e2e validation should run from this exact commit

[x] I am using an agent and I take responsibility for this PR

Assisted-by: GPT-5 via pi

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of Go-based builds in remote sandboxes by explicitly setting the Go toolchain root (`GOROOT`) for Go build steps.
  * Updated Go-related build recipes to prevent remote action failures stemming from missing Go environment configuration.
* **Documentation**
  * Added a “Lessons Learned” note explaining why dependent Go elements require an explicit `GOROOT` in remote sandboxes.
* **Chores**
  * Refreshed installation steps to write systemd presets/units and config files via a safer two-step file creation approach.

### DIFF
--- a/docs/skills/buildstream.md
+++ b/docs/skills/buildstream.md
@@ -152,3 +152,7 @@ Do not cancel a build under 120 min just because it "seems slow." Historical ran
 ### 32 fetchers is the right setting for cache.projectbluefin.io (2026-06-23)
 
 `buildstream-ci.conf` uses `fetchers: 32` (BST default is 10). With default + nvidia running simultaneously = 64 concurrent gRPC streams. The CAS server is a Hetzner AX102-U (1 Gbit/s uplink, NVMe Gen4) and can serve 64 streams comfortably. The bottleneck is network bandwidth (~125 MB/s total), not server capacity. Do not reduce fetchers without evidence of server-side saturation.
+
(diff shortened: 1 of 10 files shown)


hanthor (changes requested):
The GOROOT changes and stdin-device replacements are plausible, but this is not yet at the repository verification gate: the PR explicitly says the exact commit still needs the remote lab rebuild plus boot/e2e validation, while the repo requires the image to boot before approval. Please attach that evidence. Also document the separate no-/dev/stdin rule for remote sandboxes in the relevant skill; it is a non-obvious workaround applied across seven elements, but the current skill update only records GOROOT.

PR #1224, opened 2026-07-28
fix(lab-check): use GITHUB_TOKEN instead of MergeRaptor App token

castrojo:
Same fix as projectbluefin/bluefin#916 — align with bluefin-lts pattern, drop the App token step, use `checks: write` in workflow permissions + `${{ github.token }}` directly.


castrojo:
Superseded — updating PR #1223 instead

PR #1238, opened 2026-07-29
ci(release): align FSDK versioning and CI runners

castrojo:

This lines us up with FSDK versioning, YY.MM with all the tags we need./ 

## Testing

- [ ] `just validate` passes

## Checklist

````markdown
```verify
```
````

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -170,13 +170,12 @@ re-run the pre-flight until it is clean. Only then proceed.
 - "The stale build is for a different branch, it won't interfere" → **It uses the same runners and CAS. Cancel it.**
 - "I already cancelled one build, that's enough" → **Cancel ALL of them. Run the pre-flight again.**
 
-Independent BST **workflow runs** share the same remote executor and CAS. Never
-start a second build workflow while one is active. The four x86_64 matrix jobs inside
-one `build.yml` run are the intentional exception: they start together and are bounded
-by the BuildBox backend's four global action slots. Do not cancel matrix siblings or
-serialize them; their CAS payloads stay remote and BuildBox enforces capacity.
+Concurrent BST builds share the same `ubuntu-24.04` runner pool and the same remote CAS
+write bandwidth at `cache.projectbluefin.io:11002`. Two concurrent builds do not halve
+wall time — they more than double it and risk 6-hour timeouts with
+`Cached elements after warm: 0`.
 
-**One build workflow run, with its four coordinated variants. Field clear first.**
+**One build. Field clear first. No exceptions.**
 
 **Pre-flight is atomic:** cancel → verify clear → push/dispatch must complete in one
 uninterrupted sequence. Cancelling the active build and then not pushing the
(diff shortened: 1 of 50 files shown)

### CODE REVIEW

**Review by hanthor — CHANGES_REQUESTED** [review]
Holding this one — there's a change in here with real user-facing consequences that isn't described anywhere in the PR.

### The default image tag flips `latest` → `testing`

`include/os-release.yml`:
```diff
-  IMAGE_TAG: "latest"
+  IMAGE_TAG: "testing"
```

`elements/oci/bluefin.bst`:
```diff
-  'org.opencontainers.image.ref.name': 'ghcr.io/projectbluefin/%{image-repo}:latest'
+  'org.opencontainers.image.ref.name': 'ghcr.io/projectbluefin/%{image-repo}:%{image-tag}'
```

`org.opencontainers.image.ref.name` is baked into the image and is what `bootc` reports as the origin. If the release path doesn't override `image-tag` for stable builds, every stable Dakota image ships advertising `:testing` as its upgrade source. That's the kind of thing that's invisible until someone's `bootc upgrade` starts pulling from the wrong channel.

Please either confirm (and note in the description) that `execute-release.yml` / the stable element overrides `image-tag`, or make the default `latest` and set `testing` only in the testing path. Parameterizing the annotation via `%{image-tag}` is the right move regardless — it's the choice of default I'm questioning.

### The PR needs a description and testing evidence

Body is `## Changes` with nothing under it, and every checkbox is unticked — including `just validate`. For 50 files / +2455 lines touching release workflows, `os-release`, OCI element config, and the changelog tooling, I'd like:

- a few lines on what "align FSDK versioning" means concretely (YY.MM applied where, which tags are produced, what changes for a user reading `os-release`)
- `just validate` green
- confirmation of what the produced tag set looks like before and after, for at least one variant

### Also

The PR bundles CI runner changes with the versioning change. If the runner bump is independent, splitting it out would let it land now and keep the versioning discussion focused.

Happy to re-review quickly once the tag default is clarified — the parameterization work itself looks fine.

---

hanthor (commented):
Correcting my earlier review: **the `latest` → `testing` default is not a bug, and my concern about stable images advertising `:testing` was wrong.** I checked the export path on this branch rather than judging the two changed lines in isolation.

`just export` rewrites the OCI index annotation after checkout, from `OCI_IMAGE_TAG`:

```python
ref_name = f"ghcr.io/projectbluefin/{image_name}:{image_tag}"
for manifest in index.get("manifests", []):
    manifest.setdefault("annotations", {})["org.opencontainers.image.ref.name"] = ref_name
```

and the same recipe seds `IMAGE_TAG=` in `/usr/lib/os-release` and `image-tag` in `image-info.json` to `METADATA_TAG`. `OCI_IMAGE_TAG` defaults to `image_tag` (`BUILD_IMAGE_TAG`, default `testing`), and both `publish.yml` and `build-aarch64.yml` set it from `steps.context.outputs.export_tag`.

So the `image-tag: testing` in the `.bst` files and `IMAGE_TAG: "testing"` in `os-release.yml` are local-build defaults that CI overwrites at export time for `:next`, `:stable`, and immutable `:SHA` exports. Parameterizing the annotation via `%{image-tag}` is strictly an improvement over the previously hardcoded `:latest`, which was wrong for *every* channel. The comment above the Python block already says this — I should have read the recipe before flagging it.

My remaining asks stand and are all process, not correctness:

- a real description — "align FSDK versioning" needs a couple of sentences on what YY.MM applies to and which tags are produced
- `just validate` green (still unchecked)
- ideally the before/after tag set for one variant

Happy to approve on those; the code itself looks right. Sorry for the noise.

---

ahmedadan:
@castrojo @hanthor Can we align on what's next for this PR, should we close and try again?

castrojo:
Just redo it I'm ooo

PR #1296, opened 2026-08-07
fix(common): flatten Bluefin face icons into /usr/share/pixmaps/faces

### LINKED ISSUE #353: Parity: user account dinosaur icons are missing

castrojo:
Closes #353

## Problem

`projectbluefin/common` ships the Bluefin user-avatar art under
`system_files/bluefin/usr/share/pixmaps/faces/bluefin/` (15 JPEGs, including the
dinosaur images). `elements/bluefin/common.bst` copies `system_files/bluefin/usr/`
verbatim, so those files land at `/usr/share/pixmaps/faces/bluefin/` in the image
— confirmed by `files/fakecap-manifest.tsv:657723-657737`, which maps all 15 to
`bluefin/common.bst`.

GNOME's account-avatar chooser reads `/usr/share/pixmaps/faces` only, so the
images are present on disk but never offered to users. `ublue-os/bluefin` solves
this in `build_files/base/05-override-install.sh` (lines 29-32) by flattening
`faces/bluefin/*` over `faces/` at build time; dakota has no equivalent step.

## Change

- `elements/bluefin/common.bst` `install-commands`: move
  `%{datadir}/pixmaps/faces/bluefin/*` into `%{datadir}/pixmaps/faces/` and
  remove the now-empty subdirectory, guarded by a `-d` test.
- Same element `overlap-whitelist`: add `/usr/share/pixmaps/faces/*`. The
  flattened filenames are identical to the GNOME OS defaults from
  `core/gnome-control-center.bst` (`files/fakecap-manifest.tsv:657738-657752`),
  so the overlap is intentional and must be declared.
- `docs/skills/oci-layers.md`: record the vendor-subdirectory pattern.

## Validation

Not run — this environment has no container runtime (`podman`/`docker` absent),
so `just bst`, `just validate`, `just build`, `just lint`, and `just boot-test`
all cannot execute here. The change is unverified by build; please treat CI
`validate` + `e2e` as the gate.

What to check after a build:
```bash
just bst artifact list-contents oci/layers/bluefin.bst | grep pixmaps/faces
# expect 15 entries directly under usr/share/pixmaps/faces, no faces/bluefin/
```

Note on generated files: `files/filemap.json` and `files/fakecap-manifest.tsv`
still list the old `faces/bluefin/*` paths. Regenerating them requires
`bst artifact list-contents` against a populated local cache
(`docs/skills/ci.md`, "Generated Files"), which is not possible here. They are
refreshed by the monthly `update-filemap.yml` workflow; a maintainer can run
`workflow_dispatch` with `force_regenerate` after merge if an immediate refresh
is wanted. Stale entries are non-fatal — `fakecap-restore` skips paths that do
not exist (`files/fakecap/fakecap-restore.c:67-73`); the effect is only that the
15 avatars are unlabelled for chunkah layer assignment until regeneration.

### DIFF
--- a/docs/skills/oci-layers.md
+++ b/docs/skills/oci-layers.md
@@ -181,3 +181,22 @@ The OCI assembly script in `elements/oci/bluefin.bst` must run steps in this exa
 4. `build-oci` (assemble image)
 
 Running `ldconfig` after `build-oci` has no effect. Running `build-oci` before `dconf update` produces an image with stale/missing dconf databases.
+
+### Bluefin content shipped in a subdirectory GNOME reads non-recursively (2026-08-07)
+
+`projectbluefin/common` ships the Bluefin user-avatar art under
+`system_files/bluefin/usr/share/pixmaps/faces/bluefin/`, which
+`elements/bluefin/common.bst` copies verbatim into the image. GNOME's account
+avatar chooser scans `/usr/share/pixmaps/faces` non-recursively, so the files
+were present in the image but never visible to users (dakota#353).
+
+`ublue-os/bluefin` handles this in `build_files/base/05-override-install.sh` by
+flattening `faces/bluefin/*` over `faces/` at build time. The BST equivalent is
+a flatten step in the `install-commands` of `elements/bluefin/common.bst`, plus
+an `overlap-whitelist` entry for `/usr/share/pixmaps/faces/*` because the
+flattened filenames collide with the GNOME OS defaults from
+`core/gnome-control-center.bst`.
+
+General rule: when `common` places files in a vendor subdirectory, check whether
+the consuming component actually scans subdirectories before assuming the copy
+is sufficient.
--- a/elements/bluefin/common.bst
+++ b/elements/bluefin/common.bst
@@ -25,6 +25,9 @@ public:
     - /etc/environment
     - /etc/containers/policy.json
     - /usr/share/pixmaps/gnome-boot-logo.png
+    # Bluefin user-avatar art replaces the GNOME OS defaults shipped by
+    # core/gnome-control-center.bst (same 15 filenames).
+    - /usr/share/pixmaps/faces/*
 
 variables:
   strip-binaries: ""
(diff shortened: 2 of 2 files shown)


hanthor (approved):
Approving the substance — the diagnosis is correct and the BST translation matches the upstream precedent. One blocking mechanical issue and one edge case.

**Blocking (mechanical, not content): this PR targets `main`.** The `on-pr-opened-or-updated` gate fails for exactly that reason:

```
BASE_REF: main
ERROR: PRs must target 'testing'. Got: 'main'.
```

Heads-up before you retarget: `main` is currently ahead of `testing` by 5 commits, so flipping the base alone will pull those into the diff. (I hit the same thing on bluefin#1057, where the equivalent retarget turned a 2-file PR into a 96-file one.) Five commits is small enough that it may resolve cleanly, but check the file count after retargeting and cherry-pick onto `testing` if it balloons.

**On the change itself.** The non-recursive scan is the real root cause, and citing `05-override-install.sh` as the upstream precedent is the right justification for flattening rather than teaching GNOME to recurse. The `overlap-whitelist` entry is necessary and correctly scoped — without it the 15 filename collisions against `core/gnome-control-center.bst` would fail the build rather than silently mis-layer, so that's the safe failure direction.

The skill-doc note is the most valuable part of this PR. "When `common` places files in a vendor subdirectory, check whether the consuming component actually scans subdirectories" generalizes past this one bug, and dating it makes it auditable later.

**Edge case worth fixing while you're in there:**

```bash
if [ -d "%{install-root}%{datadir}/pixmaps/faces/bluefin" ]; then
  mv "%{install-root}%{datadir}/pixmaps/faces/bluefin/"* "%{install-root}%{datadir}/pixmaps/faces/"
  rmdir "%{install-root}%{datadir}/pixmaps/faces/bluefin"
fi
```

The `-d` guard covers a missing directory but not an *empty* one — the glob then stays literal, `mv` fails, and since BuildStream runs `install-commands` under `set -e` that aborts the build with an opaque error. This is a real possibility given the whole point is that `common` controls those contents upstream: if `common` ever ships the directory without art, dakota's build breaks rather than degrading. Cheap fix:

```bash
if compgen -G "%{install-root}%{datadir}/pixmaps/faces/bluefin/*" > /dev/null; then
```

or add `2>/dev/null || true` to the `mv` if you'd rather keep it terse.

Also worth confirming: `mv` here silently overwrites the GNOME OS defaults, which is intended, but it means a *partial* set of Bluefin faces leaves a mix of both styles in the chooser. If `common` ships all 15, that's moot — just flagging that the flatten has no all-or-nothing property.

ahmedadan:
Closed as superseded by #1302

PR #1302, opened 2026-08-07

### LINKED ISSUE #353: Parity: user account dinosaur icons are missing

mrbobbytables:
## Summary

Closes #353 — Bluefin's custom dinosaur-themed avatars never show up in the
GNOME Users account-picture picker on Dakota.

## Root cause of why #353 is still open

The fix for this was already written, reviewed, and merged in #1241 — but
#1241 was merged directly into `main`. Per `docs/workflow.md`, `main` is a
release bookmark only, fast-forwarded from `testing` by
`execute-release.yml`, and **must not be used as a PR base**. All content
PRs are required to target `testing`, the actual development trunk that
gets built and shipped.

Because #1241 landed on `main` instead of `testing`, the fix never made it
into any built/shipped image (verified: `elements/bluefin/user-avatars.bst`
exists on `upstream/main` but not on `upstream/testing`), which is why #353
is still open despite a merged 'fix'.

## Change

This PR re-applies the exact same, already-reviewed change from #1241
(cherry-picked, 3 files, no modifications) onto `testing`:

- `elements/bluefin/user-avatars.bst` — new manual element installing a
  distro dconf override.
- `files/user-avatars/07-dakota-avatar-directories` — sets
  `org.gnome.desktop.interface avatar-directories` to
  `['/usr/share/pixmaps/faces/bluefin']`, so gnome-control-center's avatar
  chooser and gnome-initial-setup's account page (which only enumerate
  `<datadir>/pixmaps/faces/` directly, no recursion) pick up Bluefin's
  dinosaur avatars from the `bluefin/` subdirectory already shipped by
  `bluefin/common.bst`.
- `elements/bluefin/deps.bst` — wires in the new element.

## Testing

- [ ] `just validate` passes

Not run in this environment — no BuildStream/podman available here (same
constraint noted in #1241 and in #1296). This is a byte-identical port of
previously merged, reviewed content; please treat CI `validate` + `e2e` on
`testing` as the gate, same as for any other PR.

## Note on #1296

#1296 (open, targets `main`) proposes an alternative approach (flattening
files at build time instead of the dconf `avatar-directories` override) and
has the same base-branch issue flagged by the bot. Maintainers may prefer
that approach instead — either way, whichever fix lands should target
`testing`, not `main`.

Closes #353

### DIFF
--- a/elements/bluefin/deps.bst
+++ b/elements/bluefin/deps.bst
@@ -23,6 +23,7 @@ depends:
   - bluefin/umotd.bst
 
   - bluefin/common.bst
+  - bluefin/user-avatars.bst
   - bluefin/just-overrides.bst
   - bluefin/zswap-config.bst
   - bluefin/swapfile.bst
--- a/elements/bluefin/user-avatars.bst
+++ b/elements/bluefin/user-avatars.bst
@@ -1,0 +1,21 @@
+kind: manual
+
+# Points GNOME's avatar chooser (gnome-control-center) and gnome-initial-setup's
+# account page at bluefin-common's dinosaur avatar directory. See the keyfile
+# for the full explanation. Fixes: projectbluefin/dakota#353
+sources:
+- kind: local
+  path: files/user-avatars
+
+build-depends:
+- freedesktop-sdk.bst:public-stacks/runtime-minimal.bst
+
+variables:
+  strip-binaries: ""
+
+config:
+  install-commands:
+  - |
+    install -Dm644 07-dakota-avatar-directories \
+      "%{install-root}%{sysconfdir}/dconf/db/distro.d/07-dakota-avatar-directories"
+  - "%{install-extra}"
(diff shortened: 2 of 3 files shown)


castrojo (approved):
LGTM. Verified:

- Base branch is `testing` — correct per docs/workflow.md (and the explanation for why #1241 never shipped, having landed on the `main` bookmark, checks out).
- Install path `%{sysconfdir}/dconf/db/distro.d/07-dakota-avatar-directories` follows the repo's existing `05-`/`06-` distro.d keyfile convention; `oci/bluefin.bst` runs `dconf update /layer/etc/dconf/db` so the override compiles into the db.
- `bluefin/common.bst` sets `system-db:distro` in `/etc/dconf/profile/user`, and the dinosaur avatars do exist at `system_files/bluefin/usr/share/pixmaps/faces/bluefin/*.jpg` in projectbluefin/common — the keyfile's `avatar-directories=['/usr/share/pixmaps/faces/bluefin']` targets the correct directory.
- CI `validate` passes.

Note: #1308 contains this exact same 3-file change plus an `avatar-audit` CI recipe. Whichever lands first, the other's fix half becomes a no-op — maintainers should pick one. I'd slightly prefer #1308 for the built-image audit, but this one is correct as-is.

PR #1308, opened 2026-08-07
test: add avatar-audit to catch #353-class dinosaur-avatar regressions

### LINKED ISSUE #353: Parity: user account dinosaur icons are missing

Danathar:
## Summary

Closes #353 — Bluefin's dinosaur-themed avatars never appear in the GNOME
Users account-picture picker on Dakota.

## Why #353 is still open

The fix was already written and merged in #1241, but it landed on `main`
(the release bookmark, fast-forwarded from `testing` and never a valid PR
base per `docs/workflow.md`) instead of `testing`, so it never shipped in a
built image. `elements/bluefin/user-avatars.bst` exists on `upstream/main`
but not on `upstream/testing`.

This PR cherry-picks that exact, already-reviewed 3-file change
(`c5828e0`, byte-identical to #1241) onto `testing`, and adds a CI check so
this class of "shipped but unreachable" bug gets caught before merge instead
of silently regressing again.

Note: #1302 independently re-applies the same cherry-pick onto `testing`.
This PR is equivalent for that part; the addition here is the `avatar-audit`
Justfile recipe below. Maintainers should pick whichever PR they prefer for
the base fix — happy to close this in favor of #1302 if the audit recipe is
cherry-picked over there instead, or vice versa.

## What's new: `just avatar-audit`

Added a `swap-audit`-style Justfile recipe, wired into `publish.yml` right
after `swap-audit`, that inspects the *built* image rather than just the
source tree:

- Bluefin avatar art actually exists under
  `/usr/share/pixmaps/faces/bluefin/*.jpg`.
- `/etc/dconf/profile/user` reads `system-db:distro` at all.
- The `07-dakota-avatar-directories` keyfile is present.
- Every directory listed in `avatar-directories` exists in the image, has
  `.jpg` faces, and is actually present in the *compiled*
  `/etc/dconf/db/distro` (catches `dconf update` ordering regressions, not
  just keyfile-not-installed).

This matters because `org.gnome.desktop.interface avatar-directories`
**replaces** GNOME's default faces dirs rather than falling back to them
(`cc-avatar-chooser.c` in gnome-control-center, `um-photo-dialog.c` in
gnome-initial-setup both try configured dirs first and only fall back to
`<datadir>/pixmaps/faces` if that yields zero faces) — so a stale, typo'd,
or dropped override doesn't degrade to stock icons, it silently empties the
picker entirely. That failure mode is invisible from `just bst artifact
list-contents` alone, which is exactly what let #353 go unnoticed even
though the art was correctly ingested the whole time.

Also documented this pattern in `docs/skills/oci-layers.md` under Lessons
Learned per the repo's self-improvement mandate — "files present in the
built image" and "files GNOME will actually discover" are different claims,
and this is the second time that gap has bitten this repo (see the FDSDK
GL-merge-symlink lesson just above it).

## Testing

- `just check-publish-workflow` and `python3 -m unittest
  scripts.test_check_publish_workflow` pass.
- `just --summary` / `just --dry-run avatar-audit` confirm the Justfile
  parses and the recipe body is well-formed.
- The audit script's inner logic was extracted and run standalone against 7
  synthetic image-root fixtures (happy path + 6 failure modes: missing
  keyfile, stale compiled db, dangling directory reference, empty target
  directory, dconf profile not reading the distro db, empty
  `avatar-directories` value) — all pass/fail as expected.
- Could not run `just validate` / `just build` / `just boot-test` / `just
  lint` in this environment (no BuildStream/podman available, same
  constraint noted in #1241, #1296, #1302); requesting CI `validate` + `e2e`
  on `testing` as the gate, and a maintainer build to confirm `avatar-audit`
  passes against the real image.

```verify
# Settings → Users → click your account picture → dinosaur avatars should
# appear in the picker (or on a fresh install, gnome-initial-setup's
# account page should offer them).
gsettings get org.gnome.desktop.interface avatar-directories
# expect: ['/usr/share/pixmaps/faces/bluefin']
ls /usr/share/pixmaps/faces/bluefin/*.jpg
```

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -190,6 +190,11 @@ jobs:
           BUILD_IMAGE_NAME: ${{ env.IMAGE_NAME }}${{ matrix.image_suffix }}
         run: just swap-audit
 
+      - name: Audit user-avatar configuration
+        env:
+          BUILD_IMAGE_NAME: ${{ env.IMAGE_NAME }}${{ matrix.image_suffix }}
+        run: just avatar-audit
+
       - name: Login to GHCR
         env:
           GH_ACTOR: ${{ github.actor }}
--- a/Justfile
+++ b/Justfile
@@ -1258,3 +1258,81 @@ swap-audit:
         [ "$STATUS" -eq 0 ] && echo "PASS: swap/zram configuration OK"
         exit "$STATUS"
         '
+
+# ── Avatar audit ─────────────────────────────────────────────────────
+# Assert the Bluefin dinosaur avatars are actually reachable by GNOME's
+# account-picture pickers. Guards against the #353 regression class, where
+# the art shipped fine but nothing pointed GNOME at it.
+#
+# Two independent ways this breaks silently, both checked here:
+#   1. common.bst stops ingesting /usr/share/pixmaps/faces/bluefin/*.jpg.
+#   2. The dconf override ships but never reaches the compiled distro db
+#      (element dropped from the layer, or `dconf update` did not run).
+#
+# org.gnome.desktop.interface avatar-directories REPLACES the default face
+# dirs rather than adding to them (gnome-control-center
+# panels/system/users/cc-avatar-chooser.c, gnome-initial-setup
+# pages/account/um-photo-dialog.c), so a stale or typo'd path yields an
+# empty picker with no fallback. Every listed dir is therefore checked for
+# existence and for actual .jpg content.
+[group('test')]
+avatar-audit:
+    #!/usr/bin/env bash
+    set -euo pipefail
+
+    # Use sudo unless already root
+    SUDO_CMD=""
+    if [ "$(id -u)" -ne 0 ]; then
+        SUDO_CMD="sudo"
+    fi
+
+    echo "==> Auditing user-avatar configuration in {{image_name}}:{{image_tag}}..."
+    $SUDO_CMD podman run --rm --pull=never \
+        "{{image_name}}:{{image_tag}}" \
+        bash -c '
+        set -uo pipefail
+        STATUS=0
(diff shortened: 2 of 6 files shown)


castrojo (approved):
LGTM — verified the fix mechanics and the audit against the repo:

**Fix portion** (identical to #1302, cherry-pick of #1241 onto the correct `testing` base):
- `elements/bluefin/user-avatars.bst` installs to `%{sysconfdir}/dconf/db/distro.d/07-dakota-avatar-directories` — consistent with the existing `05-dakota-custom-command-menu`/`06-dakota-keybindings` keyfiles, and `oci/bluefin.bst` runs `dconf update /layer/etc/dconf/db`, so the override will reach the compiled db.
- `bluefin/common.bst` writes `system-db:distro` into `/etc/dconf/profile/user`, and the source art exists at `system_files/bluefin/usr/share/pixmaps/faces/bluefin/*.jpg` in projectbluefin/common — the keyfile's `avatar-directories=['/usr/share/pixmaps/faces/bluefin']` points at the right place.

**Audit portion:**
- `avatar-audit` mirrors `swap-audit` (`--pull=never` against `{{image_name}}:{{image_tag}}`, correct use of the Justfile's `BUILD_IMAGE_NAME`/`BUILD_IMAGE_TAG` env vars).
- Placement in `publish.yml` directly after the `swap-audit` step in the matrix build job is correct — the image exists locally at that point. `check_publish_workflow.py` exists and CI is green.
- The compiled-db grep (`grep -qaF "$dir" /etc/dconf/db/distro`) is a sound gvdb check, and auditing every directory parsed out of the keyfile correctly guards the replace-not-fallback semantics of `avatar-directories`.

Note: the fix half fully overlaps #1302 (same 3 files). Either can land; if this one lands, #1302 can close. CI green.

PR #1304, opened 2026-08-07
docs(e2e-ci): spec installer post-boot assertions for fisherman e2e gaps

onlydole:
## Summary

dakota#651 is a tracking issue for e2e coverage gaps against a fisherman-driven
(the `bootc-installer` Go backend) install, filed here because issues are
disabled on `tuna-os/fisherman`. The three source fixes it tracks
(fisherman #1, fisherman #2, projectbluefin/common#385) live in other repos
Dakota doesn't own — but the *installed system* those fixes affect is still
Dakota's problem to verify, per the installer.md repo-boundary rule.

This PR adds the missing spec so the testsuite install-flow suite (or
whoever implements the actual assertions) has a single documented reference:

- **`docs/skills/e2e-ci.md`** — new "Installer Post-Boot Assertions
  (fisherman)" section with a table of all three assertions, which upstream
  fix each depends on, and an explicit rule that these belong in the
  testsuite install-flow suite, not Dakota's own generic-image boot-check
  gate (different code path — `bootc install to-disk`/`to-filesystem` via
  fisherman with LUKS vs. Dakota's `--via-loopback` raw-disk build). Also
  adds a sequencing note: land as `xfail`/skip-until-fixed against current
  fisherman, flip to blocking once each source fix ships in a nightly.
- **`docs/skills/installer.md`** — cross-link from Lessons Learned back to
  the new e2e-ci.md section.

No workflow/Justfile changes — this is documentation only, scoping the work
so the actual assertion implementation (which lives in
`projectbluefin/testsuite`, per this repo's "Dakota does not own any
behave/step code" rule) has a clear, agreed spec to build against.

Closes #651

## Verification

- Docs-only change; no build/lint required for this diff.
- Cross-checked against `docs/skills/e2e-ci.md`'s existing Testsuite Rule and
  `docs/skills/installer.md`'s Dakota vs Dakota-ISO boundary table to keep
  the new section consistent with established conventions.

- [x] I am using an agent and I take responsibility for this PR

Issue #651, opened 2026-06-01
[fisherman] e2e: add post-boot assertions for UEFI entries, Flatpak exclusion, and LUKS cmdline format


hanthor:
```
DAKOTA 🦖  ·  issue pipeline
─────────────────────────────────────────────────
  ▶  filed      report received
  ·  approved   —
  ·  queued     —
  ·  claimed    —
  ·  done       —
─────────────────────────────────────────────────
  report:       missing     ·  confirms: 0
  area:         —         ·  priority: —
  next action:  same bug? ujust confirm 651
```

---Tracking issue for e2e test gaps in fisherman (the installer backend). Issues are disabled on the fisherman repo; tracking here.

## Source issues
- fisherman #1: exclude installer Flatpaks from the installed system
- fisherman #2: mount host `/sys` in podman so `efibootmgr` can write UEFI variables
- projectbluefin/common#385: fix `rd.luks.name=` parsing in `luks-tpm2-autounlock`

### 1. UEFI boot entry written to firmware (fisherman #2)

After a to-filesystem install, `efibootmgr -v` on the booted system must show the installer-written entry. Without the `/sys/firmware/efi/efivars` bind-mount in the `podman run` invocation, efibootmgr can't reach host UEFI variables.

### 2. Installer Flatpaks excluded from target (fisherman #1)

`CopyFlatpaks` copies the system flatpak store to the installed system. On the live ISO the tuna-installer app is in the system store and must not appear on the installed target.

### 3. LUKS cmdline UUID parseable with rd.luks.name= format (common#385)

After a LUKS install, `/proc/cmdline` should contain a parseable LUKS UUID using either `rd.luks.uuid=` or `rd.luks.name=` format. Verifies the `luks-tpm2-autounlock` fix will work on the installed system.

```verify
# After fisherman fix merges and a nightly ships:
flatpak list --system --app | grep org.bootcinstaller
# Expect: no output
efibootmgr -v
# Expect: BootCurrent + Boot entries present
```

castrojo:
## Verification 2026-08-08: the assertions this issue asks for have been implemented and merged — in `projectbluefin/testsuite` PR #697; dakota-side spec is in open PR #1304

State of each requested assertion, checked against `projectbluefin/testsuite@main` (tip `ef6afa7`):

| This issue's ask | Where it landed |
|---|---|
[...]

PR #1316, opened 2026-08-08
ci(e2e): expose installer suite on workflow_dispatch (fisherman post-boot assertions)

hanthor:
## Summary

[dakota#651](https://github.com/projectbluefin/dakota/issues/651) tracks e2e coverage gaps against a fisherman (`bootc-installer` Go backend) to-filesystem install. The post-boot assertions themselves are implemented and merged in `projectbluefin/testsuite` (#697, `tests/installer/features/installer_post_boot.feature`), and the dakota-side spec lives in open PR #1304 — but as noted in the issue thread, **no dakota workflow requests the `installer` suite**, so the assertions can't be run from this repo at all.

This PR closes that gap on the only safe surface: the **workflow_dispatch-only** `e2e.yml`.

- Adds `installer` to the `suites` dispatch input (the testsuite reusable workflow already registers and supports it).
- Documents the suite in the workflow header: what the three post-boot assertions are, where they're implemented (testsuite #697), the tracking issue/spec location (dakota#651, `docs/skills/e2e-ci.md`), and the skip semantics — UEFI/LUKS scenarios skip unless the target is a fisherman-installed UEFI+LUKS system, so on the standard direct-kernel-boot lane only the Flatpak-exclusion assertion runs.

**Explicitly out of scope (per the #1304 sequencing rule):** wiring `installer` into `publish-smoke.yml` or any automated pipeline. Those assertions verify upstream fisherman/common source fixes that haven't shipped yet, so they must land skip-until-fixed — a maintainer sequencing decision, not something to bundle here. This change is dispatch-only and cannot turn any automated CI red.

Refs #651 (tracking); complements #1304 (spec docs).

## Verification

- Workflow-only change: no build/lint required.
- `installer` suite verified present in `projectbluefin/testsuite`'s reusable `e2e.yml` (suites input + SSH-only runner path).
- YAML is a comment block + one quoted scalar change; parsed structure unchanged.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -7,6 +7,16 @@
 #
 # Only runs on workflow_dispatch — PRs do not publish a testing build first,
 # so running smoke here would test a stale :testing image, not the PR's code.
+#
+# The `installer` suite runs the fisherman post-boot assertions (UEFI boot
+# entry via efibootmgr, installer Flatpak exclusion, LUKS rd.luks.name=
+# cmdline format) over SSH against the target image. Implemented in
+# projectbluefin/testsuite (tests/installer/features/installer_post_boot.feature);
+# tracked by projectbluefin/dakota#651 with the spec in docs/skills/e2e-ci.md
+# -> "Installer Post-Boot Assertions (fisherman)". The UEFI/LUKS scenarios
+# skip unless the target is a fisherman-installed UEFI+LUKS system, so on
+# the standard direct-kernel-boot lane only the Flatpak-exclusion assertion
+# runs; the suite is meaningful against a to-filesystem install.
 
 name: E2E Tests — GNOME 50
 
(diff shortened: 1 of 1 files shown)


castrojo (approved):
LGTM. Verified against the repo:

- The `installer` suite is real: `projectbluefin/testsuite`'s reusable `e2e.yml` lists it in the `suites` input ("smoke,developer,dx,software,vanilla-gnome,bazzite,common,lifecycle,installer,kde-smoke"), so the new dispatch description matches what the callee accepts.
- Change is correctly scoped to the `workflow_dispatch`-only `e2e.yml` — no automated pipeline can go red from this, consistent with the #1304 sequencing rule (assertions must land skip-until-fixed).
- Header comment accurately reflects dakota#651 and the skip semantics for UEFI/LUKS on the direct-kernel-boot lane.
- CI green (validate + triage pass).

Non-blocking: the comment says fisherman issues are tracked via dakota#651 because issues are disabled on tuna-os/fisherman — that repo actually has issues enabled, but the dakota-side tracking reference here is accurate regardless.

PR #1334, opened 2026-08-13

### LINKED ISSUE #353: Parity: user account dinosaur icons are missing

kylerankin:
## Problem

Bluefin ships custom dinosaur-themed user avatars under `/usr/share/pixmaps/faces/bluefin/` (a subdirectory of the standard faces dir). However, GNOME's avatar chooser (in gnome-control-center) and gnome-initial-setup's account page only enumerate files directly inside `<datadir>/pixmaps/faces/` — they never recurse into subdirectories.

As a result, both UIs silently fall back to Fedora/GNOME's stock (near-empty) faces dir, and Bluefin's dinosaur avatars never appear on Dakota.

## Fix

Set `org.gnome.desktop.interface avatar-directories` via a distro dconf override to point both UIs at the `bluefin/` subdirectory explicitly.

Closes #353

### DIFF
--- a/elements/bluefin/deps.bst
+++ b/elements/bluefin/deps.bst
@@ -22,6 +22,7 @@ depends:
   - bluefin/umotd.bst
 
   - bluefin/common.bst
+  - bluefin/user-avatars.bst
   - bluefin/just-overrides.bst
   - bluefin/firstboot-date.bst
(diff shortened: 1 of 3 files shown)


ahmedadan:
Closed as superseded by #1302

PR #1335, opened 2026-08-13
fix: restore Bluefin dinosaur avatars in GNOME user-account picker (#353)

### LINKED ISSUE #353: Parity: user account dinosaur icons are missing

kylerankin:
## Problem

gnome-control-center's avatar chooser (cc-avatar-chooser.c) and gnome-initial-setup's account page only enumerate files directly inside `<datadir>/pixmaps/faces/` — they never recurse into subdirectories.

bluefin-common ships its dinosaur-themed avatars in a `bluefin/` category subdirectory (`/usr/share/pixmaps/faces/bluefin/*.jpg`), so both UIs silently fall back to Fedora/GNOME's near-empty stock faces dir and Bluefin's avatars never appear.

## Fix

Set `org.gnome.desktop.interface avatar-directories` via a distro dconf override to point both UIs at the bluefin/ subdirectory, matching the existing dconf-override pattern used elsewhere in this repo.

### Files changed

- `elements/bluefin/user-avatars.bst` — new BST element (manual kind) that installs the dconf override
- `files/user-avatars/07-dakota-avatar-directories` — dconf keyfile setting `avatar-directories=[/usr/share/pixmaps/faces/bluefin]`
- `elements/bluefin/deps.bst` — added `bluefin/user-avatars.bst` dependency after `bluefin/common.bst`

Closes #353

### DIFF
--- a/elements/bluefin/deps.bst
+++ b/elements/bluefin/deps.bst
@@ -23,6 +23,7 @@ depends:
   - bluefin/umotd.bst
 
   - bluefin/common.bst
+  - bluefin/user-avatars.bst
   - bluefin/just-overrides.bst
   - bluefin/zswap-config.bst
   - bluefin/swapfile.bst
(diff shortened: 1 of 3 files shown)


ahmedadan:
Closed as superseded by #1302

PR #1467, opened 2026-09-01
fix(ci): open monthly filemap PRs against testing, not main

ahmedadan:
## Summary

The monthly filemap regeneration workflow hardcoded `--base main` in its `gh pr create` call, a leftover from when main was the integration branch, so every generated PR targeted main and had to be retargeted by hand (#1460 was the latest). One line: the base is now testing, matching where all integration happens.

Related: #1460


hanthor (approved):
One-line fix: the monthly filemap regeneration workflow's `gh pr create --base main` is changed to `--base testing`, matching the branch where dakota integration actually happens now. Confirmed against the linked evidence (#1460 is the exact PR that had to be retargeted by hand because of this). `validate` and `ensure-labels` checks are green. Correct, minimal, well-justified.

---

PR #1547, opened 2026-09-10
feat(hardware): package tuxedo-drivers kernel modules and quirks

castrojo:
OK so this is 600k ondisk and they only load if you need it. I want to see how doable this is, I suspect easy with hive. Tuxedo is the only one in Europe we've had contact with, they sponsored the linuxapp summit, and we had lots of great convos. 

They want to get to the right place with upstream kernels but it's a long complicated thing and the new people are stuck fixing it. So let's give them all a kickass OS. 😄 

I'm thinking roll with this for a while but not make a habit of it. I think knowing that they intend to get better is a good indicator.

### Summary
- Packages `tuxedo-drivers` as an out-of-tree kernel module element built against Dakota's pinned kernel headers (`freedesktop-sdk.bst:components/linux.bst`).
- Stages all 27 kernel modules (`tuxedo_keyboard`, `tuxedo_io`, `clevo_acpi`, `clevo_wmi`, `uniwill_wmi`, `ite_829x`, `tuxedo_nb04_*`, `tuxedo_nb05_*`, etc.).
- Installs upstream hardware quirks and udev rules (hwdb remappings for Fn keys and accelerometer orientation matrix, modprobe conflicts blacklist).
- Includes the element in `elements/oci/layers/bluefin-stack.bst` with layer integration command `depmod -a $KVER`.
- Updates `.agents/skills/dakota-buildstream/SKILL.md` per the factory two-output rule.

Assisted-by: Pi

### DIFF
--- a/.agents/skills/dakota-buildstream/SKILL.md
+++ b/.agents/skills/dakota-buildstream/SKILL.md
@@ -46,6 +46,7 @@ Never translate RPM, DNF, COPR, or Containerfile workflows into this repository.
   ```bash
   python3 files/scripts/generate_cargo_sources.py path/to/Cargo.lock
   ```
+- **Out-of-Tree Kernel Modules**: Build against `/usr/lib/modules/$KVER/build` provided by `freedesktop-sdk.bst:components/linux.bst`. Defer `depmod` from the driver element's `install-commands` to the composed layer stack's `integration-commands` (`depmod -a "$KVER"`).
 
 ## Common Rationalizations
 
--- a/elements/bluefin/tuxedo-drivers.bst
+++ b/elements/bluefin/tuxedo-drivers.bst
@@ -1,0 +1,50 @@
+kind: manual
+description: |
+  Out-of-tree kernel modules and udev hardware quirks for TUXEDO laptops.
+  Provides platform drivers, keyboard backlight, fan and power sensors,
+  and /dev/tuxedo_io for Clevo and Uniwill/Tongfang ODM chassis.
+
+build-depends:
+- freedesktop-sdk.bst:bootstrap/coreutils.bst
+- freedesktop-sdk.bst:bootstrap/bash.bst
+- freedesktop-sdk.bst:components/gcc-base.bst
+- freedesktop-sdk.bst:bootstrap/binutils.bst
+- freedesktop-sdk.bst:components/make.bst
+- freedesktop-sdk.bst:components/linux.bst
+# Kernel's objtool is dynamically linked against libxxhash.so.0
+- freedesktop-sdk.bst:components/xxhash.bst
+
+depends:
+- freedesktop-sdk.bst:public-stacks/runtime-minimal.bst
+
+sources:
+- kind: git_repo
+  url: gitlab:tuxedocomputers/development/packages/tuxedo-drivers.git
+  track: main
+  ref: 2c6bf54075fb38a7fdbefc560734281984bf65bc
+
+config:
+  build-commands:
+  - |
+    set -euo pipefail
+    KVER=$(ls /usr/lib/modules | head -n1)
+    if [ -z "$KVER" ] || [ ! -d "/usr/lib/modules/$KVER/build" ]; then
+        echo "FATAL: no kernel build tree found under /usr/lib/modules" >&2
+        exit 1
+    fi
+    echo "Building TUXEDO kernel modules against $KVER"
+    make -j${BST_BUILD_THREADS:-1} -C "/usr/lib/modules/$KVER/build" M="$PWD" modules
+
+  install-commands:
+  # depmod is intentionally deferred to OCI layer composition time
+  - |
+    set -euo pipefail
+    KVER=$(ls /usr/lib/modules | head -n1)
+    make -j1 -C "/usr/lib/modules/$KVER/build" M="$PWD" modules_install \
+        INSTALL_MOD_PATH="%{install-root}%{prefix}"
+  - |
+    set -euo pipefail
+    # Install hardware quirks, udev rules, and modprobe configuration
+    if [ -d files/usr ]; then
+        cp -r files/usr/* "%{install-root}%{prefix}/"
+    fi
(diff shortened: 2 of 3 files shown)


hanthor (commented):
Reviewing as a draft, so this is a COMMENT rather than a block. Good news first, then four things I would fix before marking it ready.

## The red `validate` is stale, not a real failure

Both failing runs are against base `92cc4ea`, and the failure is the ChairLift runner-defaults issue:

```
FAIL - upgrade is independent of migration
error: recipe `test-chairlift-migration` failed on line 112 with exit code 1
error: recipe `check-publish-workflow` failed on line 87 with exit code 1
```

That was fixed by #1550 (`fix(ci): isolate ChairLift tests from runner defaults`), which is already on `testing`. Merging this branch onto current `testing` (`a27b307`) and running the gate locally is clean:

```
$ git merge --no-gpg-sign origin/pr/1547 && just check-publish-workflow
...
All ChairLift migration tests passed.
EXIT=0
```

A rebase/retarget will turn the check green. (Note this is only `check-publish-workflow`; I could not run the `just bst show` steps locally as there is no BuildStream in this environment.)

## 1. `ls /usr/lib/modules | head -n1` is not a safe way to pick the kernel

It appears three times — twice in `tuxedo-drivers.bst`, once in `bluefin-stack.bst` integration-commands:

```bash
KVER=$(ls /usr/lib/modules | head -n1)
```

Two problems. It takes the lexicographically first directory, which silently picks the wrong kernel the moment more than one is staged, and `ls` is not guaranteed to be the kernel you built against. Second, this runs under `set -euo pipefail`; `ls` writing into a pipe that `head` has already closed can return non-zero and abort the recipe for an unrelated reason.

The repo already has a single source of truth for this — `scripts/test_gen_filemap.py` derives the pinned version straight from `elements/core/linux-fdsdk.bst`:

```python
ref = re.search(r"^  ref: v([0-9][^-\s]*)-", element_path.read_text(), re.M)
```

Deriving `KVER` the same way, or at minimum failing loudly when `ls /usr/lib/modules` returns more than one entry, would make this deterministic.

## 2. The udev hwdb quirks are installed but never compiled, so they do nothing

Upstream ships hwdb entries at the pinned ref:

```
$ curl .../tuxedo-drivers/repository/tree?ref=2c6bf540...&path=files&recursive=true
blob files/usr/lib/udev/hwdb.d/61-keyboard-tuxedo.hwdb
blob files/usr/lib/udev/hwdb.d/61-sensor-tuxedo.hwdb
blob files/usr/lib/modprobe.d/tuxedo-drivers-backlist-upstream-conflicts.conf
blob files/usr/lib/udev/rules.d/99-tuxedo-*.rules   (7 files)
```

`cp -r files/usr/*` puts the `.hwdb` files in place, and `bluefin-stack.bst` integration re-runs `depmod -a`, but nothing regenerates `/usr/lib/udev/hwdb.bin`. `hwdb.d` fragments are inert until `systemd-hwdb update` compiles them, and there is no such call anywhere in the tree:

```
$ grep -rn "systemd-hwdb\|hwdb" --include=*.bst --include=*.yml elements/ include/
(no matches)
```

So the keyboard-backlight and sensor quirks — half of what the PR title promises — will not take effect on a built image. The `.rules` and `modprobe.d` files are fine; only the hwdb half is affected. Adding `systemd-hwdb update` next to the `depmod -a` in integration-commands should cover it.

## 3. `depmod` is described as deferred but is not actually suppressed

```yaml
  install-commands:
  # depmod is intentionally deferred to OCI layer composition time
  - |
    make -j1 -C "/usr/lib/modules/$KVER/build" M="$PWD" modules_install \
        INSTALL_MOD_PATH="%{install-root}%{prefix}"
```

Kbuild's `modules_install` runs its own depmod pass at the end; the comment does not stop it. Either it runs and bakes a `modules.dep` generated from an incomplete tree into the element (later overwritten by the layer's `depmod -a`, so mostly wasted work), or it fails and takes the build with it. If the intent is genuinely to defer, pass `DEPMOD=/bin/true` (or `depmod=true`) on that `make` line so the comment and the behaviour agree.

## 4. No architecture guard, and `oci/bluefin.bst` is built for aarch64

`bluefin-stack.bst` is a dependency of `oci/layers/bluefin.bst`, and `.github/workflows/build-aarch64.yml` line 113 runs:

```
run: just bst build oci/bluefin.bst
```

So this element will be built on aarch64, where TUXEDO's Clevo/Uniwill x86 platform drivers have no hardware to target. Sibling elements gate on arch already (`elements/bluefin/gum.bst`, `tailscale.bst` both use `(?): - arch == "x86_64":`). Worth applying the same conditional here rather than spending aarch64 build time on modules that cannot load.

## Minor

The `.agents/skills/dakota-buildstream/SKILL.md` line is a sensible addition and is consistent with what the element does, though it is not mentioned in the PR description — worth a line in the summary since it changes guidance for future work, not just this element.

---

castrojo:
Rebased onto `origin/testing` and addressed review feedback:

1. **Conflict Resolution**: Kept current testing base changes in `bluefin-stack.bst` (`gnome-build-meta.bst:oci/initramfs/image.bst`) and adapted build-depends to freedesktop-sdk 26.08 (`components/_private/gcc-base.bst`).
2. **Deterministic Kernel Selection**: Replaced `ls /usr/lib/modules | head -n1` across `tuxedo-drivers.bst` and `bluefin-stack.bst` with explicit directory iteration that detects exactly one kernel and aborts with a fatal error if multiple directories exist.
3. **Hardware Quirks Compilation**: Added `systemd-hwdb update` to `bluefin-stack.bst` integration commands following `depmod -a` so udev hwdb quirks for keyboard backlight and sensors are compiled into `/usr/lib/udev/hwdb.bin`.
4. **Depmod Deferral**: Passed `DEPMOD=/bin/true` to `make modules_install` in `tuxedo-drivers.bst` so depmod is genuinely deferred to OCI layer composition time.
5. **Architecture Guard**: Added conditional dependency `arch == "x86_64"` in `bluefin-stack.bst` so `tuxedo-drivers.bst` is only included on x86_64 targets and skipped on aarch64 builds.

Local validation ran and passed: `just validate`, `just bst show --deps all oci/bluefin.bst` (both x86_64 and with `-o arch aarch64`), and `just check-publish-workflow`.

PR #1561, opened 2026-09-12
feat(countme): replace the local reporter with the shared bluefin-countme

castrojo:
Dakota carried its own countme client that reported to Fedora — `/usr/libexec/dakota-countme` sends to `mirrors.fedoraproject.org` — while the timer triggering it already carried `Documentation=https://countme.projectbluefin.io`.

## Replaced, not supplemented

Dakota now carries no countme code of its own. The unit and its helper come from projectbluefin/common through `bluefin/common.bst`, which already does:

```
cp -r system_files/shared/usr/./ "%{install-root}%{prefix}/"
```

so this image counts through exactly the same `bluefin-countme` as every other image.

Removed:

- `files/countme/dakota-countme` — the Fedora reporter
- `files/countme/dakota-countme.service` / `.timer`

`elements/bluefin/countme.bst` keeps only the enablement symlink, because this image enables units through `timers.target.wants` rather than by running `systemctl preset` — the same mechanism `alsa-utils.bst` uses.

Net: 6 insertions, 99 deletions.

## Depends on

projectbluefin/common#1112, which adds the shared reporter. The `bluefin/common.bst` ref needs bumping to a commit that includes it, otherwise the `bluefin-countme.timer` symlink dangles. That needs `bst track` to pin.


Danathar (commented):
This is the right shape for what it wants to do, but it cannot land yet. Three reasons, in order of how much they matter.

**1. It depends on common#1112, which is still open.** That PR is what adds `bluefin-countme.service` and `bluefin-countme.timer` to common, and it still has changes requested (the `DynamicUser=yes` unit cannot run `bootc status`). At this head `elements/bluefin/common.bst` pins `v2026.08-105-g56d81ef`, and `testing` is on `v2026.08-112-g8df20e8`. Neither contains those unit files. I checked the common tree at both refs and at the #1112 head: the units only exist on #1112's branch.

So if this merged today, `countme.bst` would ship only the symlink `timers.target.wants/bluefin-countme.timer -> ../bluefin-countme.timer` with nothing behind it. The build would not fail (`ln -sf` does not check that its target exists). Dakota would stop pinging Fedora, and nothing would ping countme.projectbluefin.io either. That is a silent loss of all counting.

**2. The reverse order is also a problem.** Today `countme.bst` installs `/usr/lib/systemd/system/bluefin-countme.service` and `.timer` from `files/countme/`, and `common.bst` copies `system_files/shared/usr/` wholesale. Once the common ref is bumped past #1112 without this PR, both elements install the same two paths. `project.conf` sets no `fatal-warnings`, so BuildStream only warns about the overlap and whichever stages last wins. This is the unit-name collision I raised on common#1112.

**3. It targets `main`.** The `on-pr-opened-or-updated` check fails with "PRs must target 'testing'", and the bot has asked for the retarget twice. The branch was cut from `testing` (it carries #1517 through #1558), so retargeting also shrinks the GitHub diff from 60 files down to the 4 that are actually yours.

What I think the landing order is: once common#1112 merges, bump the `common.bst` ref in this same PR (`bst source track bluefin/common.bst`), retarget to `testing`, and I will approve. The removal of `files/countme/*` and the symlink-only `countme.bst` look correct as written. Keeping the static `timers.target.wants` link is the right call, since presets only apply on first boot and existing installs need the timer enabled on upgrade.

Two smaller things, neither blocking:
- This drops the Fedora "Dakota" ping entirely (the first commit kept both, the second removed it), so Dakota disappears from the Fedora countme dataset. I assume that is intended; worth a line in the body.
- The shared reporter starts a fresh epoch under `/var/lib/bluefin-countme`, so every existing install reports as a week-old install for a while. If that matters, common could seed from `/var/lib/dakota-countme-epoch` when it exists.

The `Lint workflow bash` failure is shellcheck on `.github/workflows/boot-test-aarch64.yml`, which this PR does not touch; it is pre-existing on the branch base.

PR #1571, opened 2026-09-15
feat(ghostty): disable resize overlay by default to prevent Wayland loop

castrojo:

On GNOME/Wayland, Ghostty's default resize overlay can trigger an infinite resize feedback loop with Mutter/compositor configure events when windows are resized or tiled. This results in the dimensions label flashing continuously and freezes window input. While Bluefin configures `resize-overlay = never` in `/etc/skel`, existing users and sessions without skel overrides run into this bug under the upstream default.

---

## Changes

- Add `patches/ghostty/0001-disable-resize-overlay-by-default.patch` to set `@"resize-overlay": ResizeOverlay = .never` in `src/config/Config.zig`.
- Wire the patch into `elements/bluefin/ghostty.bst`.
- Document the configuration default in `docs/build.md`.

## Testing

- [x] `just validate` passes

## Checklist

- [x] New BST elements wired into `deps.bst` (not a new element, wired patch into existing `ghostty.bst`)
- [x] Patches in `patches/` regenerated if junction refs changed


Danathar (approved):
Applied the patch against a fresh `ghostty-1.3.1.tar.gz` and it lands on `src/config/Config.zig` (hunk succeeds with fuzz 1, offset 1 — the target is at line 2281 in the tarball, not 2277). GNU patch accepts that, so the build should be fine, but if you touch this again it's worth regenerating the hunk with `diff -u` so it applies at zero fuzz.

Rationale is sound: the `/etc/skel` override only reaches new users, the compiled-in default reaches everyone, and `resize-overlay = after-first`/`always` still works for anyone who wants it back.

PR #1577, opened 2026-09-16
feat(review): add native host runtime dependencies

### LINKED ISSUE #1576: feat: ship container isolation tools by default
## Summary

Dakota does not currently provide the host runtime needed by the native Bluefin Review workflow:

- `crun-krun` / `krun` with `libkrun` and `libkrunfw`
- `fuse2fs` from `e2fsprogs`
- `gocryptfs`

The downstream Dudley Dakota image currently has to extract these files from a disposable Fedora builder. That workaround cannot be carried into Dakota because Dakota is assembled with BuildStream and has no RPM or DNF installation phase.

## Evidence

- `elements/bluefin/deps.bst` imports the freedesktop-sdk package graph.
- The pinned freedesktop-sdk 25.08.16 `components/e2fsprogs.bst` sets `--disable-fuse2fs`.
- The pinned freedesktop-sdk graph contains `crun.bst` but no libkrun/crun-krun element, and contains no gocryptfs element.
- Bluefin Review invokes the krun OCI runtime and checks for `fuse2fs` and `gocryptfs` on the host.

## Requested outcome

Add these tools through Dakota's native BuildStream graph, preserving the existing freedesktop-sdk and GNOME Build Meta junction architecture. The resulting image should expose the expected runtime binaries and libraries, with a focused graph or artifact validation proving the package closure.

This issue is intended to track the upstream fix for downstream parity work.

joshyorko:
Closes #1576

This draft adds the native BuildStream elements for the Bluefin Review runtime contract:

- adds Dakota's `freedesktop-sdk.bst:components/e2fsprogs.bst` dependency;
- patches the pinned freedesktop-sdk junction at its boundary to enable `fuse2fs`.
- builds `libkrun` from its pinned upstream source;
- builds `crun` with `--with-libkrun` and installs the `krun` runtime alias;
- builds `gocryptfs` from its pinned upstream source.

The patch preserves freedesktop-sdk ownership of the package element and keeps the change in Dakota's existing `patch_queue` architecture.

The Rust and Go elements use Dakota's existing BuildStream toolchain and offline dependency policy; their full dependency closure is resolved by the corresponding SDK toolchain elements and should be exercised by CI's source fetch/build jobs.

Validation:

- `git diff --check`
- `just bst show --deps all oci/bluefin.bst` (exit 0; graph resolved, remote artifact fetches remain required for a full build)
- `just patch-drift-check` (exit 0 after adding the patch manifest entry)

The focused `gocryptfs` build reaches compilation but fails because its upstream
Go modules are not declared as BuildStream `go_module` sources and the sandbox
correctly has `GOPROXY=off`. A second focused `libkrun` build was allowed to
pull SDK artifacts for two minutes but was terminated before source compilation;
the initial SDK pull is larger than the local execution window. These are
reviewable follow-up requirements for completing the source closures.


hanthor (changes requested):
Thanks for this, and thank you especially for the honesty in the PR body — saying plainly that the `gocryptfs` build fails on `GOPROXY=off` and that the `libkrun` build was terminated before it compiled anything is far more useful to a reviewer than a green checkmark would have been. I want to start with what you got right, because the shape of the change is good.

Patching `e2fsprogs` at the junction boundary rather than forking the element is the correct call. You kept freedesktop-sdk's ownership of `components/e2fsprogs.bst`, put the change in the existing `patches/freedesktop-sdk` queue, and registered it in `patches/freedesktop-sdk.manifest.json` so `just patch-drift-check` covers it. I verified the manifest hash is the real one and not a placeholder:

```
$ sha256sum 0008-e2fsprogs-enable-fuse2fs.patch
21bc72d9e306e89e5c135e0e1545538a313ffbf56d5580f8aca88aa3e7d85b79
```

which matches the manifest entry exactly. Keeping `crun` under a distinct name (`crun-krun`, with `krun` as the runtime alias) rather than shadowing a system `crun` is also right, and I confirmed there is no pre-existing `crun`/`krun`/`gocryptfs` element to collide with (`git ls-tree -r --name-only origin/testing | grep -i "crun\|krun\|gocryptfs"` is empty).

There are two blockers.

**1. The `e2fsprogs` patch does not apply to the freedesktop-sdk that `testing` is now on.**

`testing` moved to freedesktop-sdk 26.08 after you opened this (`5005f85 feat: GNOME 51 and freedesktop-sdk 26.08 on testing (#1579)`). I fetched the real upstream `elements/components/e2fsprogs.bst` at each ref and ran your patch against it:

```
$ git apply --check 0008-e2fsprogs-enable-fuse2fs.patch
freedesktop-sdk-25.08.16       APPLIES
freedesktop-sdk-25.08.17       APPLIES
freedesktop-sdk-26.08.0        error: patch failed: elements/components/e2fsprogs.bst:11
                               error: elements/components/e2fsprogs.bst: patch does not apply
```

The cause is that your hunk's context has `license-files-extra: NOTICE` immediately followed by `conf-local: >-`, which is true at 25.08 but not at 26.08 — upstream inserted an explanatory comment between them, and added an SPDX header that shifts every line number:

```
$ curl -s .../freedesktop-sdk-26.08.0/elements/components/e2fsprogs.bst | sed -n '13,20p'
variables:
  license-files-extra: NOTICE
  # The `--disable-libuuid` and `--disable-libblkid` options are disabling the
  # *bundled* versions and enabling use of the system versions of these libs.
  conf-local: >-
    --enable-elf-shlibs
    --disable-fuse2fs
```

The `--disable-fuse2fs` line you are targeting is still there, so the intent is still valid — the patch just needs regenerating against 26.08.0. Please regenerate it from the real upstream file rather than hand-writing the hunk; the `From 0000000...`, `index 1234567..89abcde` and `josh@example.invalid` placeholders in the current file are a hint that the context was written by hand, and hand-written context is exactly what breaks silently on a base bump like this one.

The rebase is needed regardless — `mergeable_state` is `dirty`, and the conflict is the manifest:

```
$ git merge-tree --write-tree origin/testing <pr-head>
CONFLICT (content): Merge conflict in patches/freedesktop-sdk.manifest.json
```

which also means the manifest's `"gnome-build-meta-sha": "d78f4981a3321e07ae4c62d4a7761de46ac8172b"` (a gnome-50 SHA) needs to become the gnome-51 one.

**2. `libkrun.bst` has the same missing-source-closure defect as `gocryptfs.bst`, and it was never actually built.**

You flagged the `gocryptfs` failure yourself: `GOPROXY: off` in the sandbox plus no `go_module` sources means the module graph cannot resolve. That is correct and the fix is to declare the Go modules as BuildStream sources.

`libkrun.bst` has the structurally identical problem and it did not surface only because the build was killed during the SDK pull, before compilation started. It is a Rust project declared with a bare `git_repo` source and nothing else:

```yaml
kind: manual
sources:
- kind: git_repo
  url: github:libkrun/libkrun.git
  track: v1.19.0
  ref: v1.19.0-0-g9932c4b59d8f891e60c6aba20d22ebb99ceaa8e2
build-depends:
- freedesktop-sdk.bst:components/rust.bst
...
config:
  build-commands:
  - make BLK=1 NET=1 GPU=1 SND=1
```

There is no `kind: cargo2` block, so no crate in libkrun's `Cargo.lock` is declared as a source, and `make` will invoke `cargo` against a network the sandbox does not have. Dakota already has the pattern you need, in the element the sibling PR #1574 is bumping: `elements/gnomeos-deps/bootc.bst` carries a full `cargo2` ref list — I verified that one covers all 415 registry crates plus 9 git crates from upstream's lockfile, with matching checksums. That is the bar a Rust element has to clear here, and it is why `--with-libkrun` in `crun-krun.bst` cannot be trusted yet either: `crun-krun` depends on `bluefin/libkrun.bst`, so it cannot build until libkrun does.

`just bst show --deps all oci/bluefin.bst` exiting 0 is a real signal but a narrow one — it proves the element graph resolves, not that any of the three new elements can be fetched or built offline. Please don't take that as validation of the source closures.

Smaller notes, not blockers:

- `libkrun.bst`'s install commands hardcode `libkrun.so.1.19.0`. That matches the pinned tag, so it is fine today, but it will break silently on the next `bst source track` of that element. A `%{}`-substituted version variable, or globbing the built soname, would be more durable.
- `crun-krun.bst` lists `autoconf`/`automake`/`libtool` individually rather than depending on `public-stacks/buildsystem-autotools.bst` the way `libkrun.bst` uses `buildsystem-make.bst`. Worth making consistent.
- Please clarify in the description whether `--enable-fuse2fs` actually installs a `fuse2fs` binary into the composed runtime, and whether `components/e2fsprogs.bst` needs a `fuse3` runtime dependency added alongside the flag. The flag flip alone does not guarantee the tool lands where Bluefin Review looks for it.

Concretely, to unblock: rebase onto current `testing`, regenerate `0008-...patch` from freedesktop-sdk 26.08.0 with its real hashes and refresh the manifest entry, add `cargo2` sources to `libkrun.bst` and `go_module` sources to `gocryptfs.bst`, and then post a build of at least `bluefin/libkrun.bst` and `bluefin/gocryptfs.bst` that reaches completion. Happy to re-review as soon as those are in — the design decisions underneath this are sound and I would like to see it land.

---

castrojo:
Thanks for working on this! Before this can land, there is a merge conflict and several build/packaging blockers that need to be resolved:

### 1. Rebase & Manifest Collision
- **Merge conflict in `patches/freedesktop-sdk.manifest.json`**: PR #1577 branched off commit `93579d1` before GNOME 51 and freedesktop-sdk 26.08 landed on `testing` (#1579). In that upgrade, old downstream patches `0001`–`0007` were dropped upstream, leaving only `0001-project.conf-Add-GNOME-CAS-servers.patch`.
- **Action**: Rebase onto `upstream/testing`, renumber `0008-e2fsprogs-enable-fuse2fs.patch` to `0002-e2fsprogs-enable-fuse2fs.patch`, run `just patch-sync` to update the manifest hashes, and verify with `just patch-drift-check`.
- Also resolve context changes in `elements/bluefin/deps.bst`.

---

### 2. `libkrun` Packaging & Runtime Dependencies
- **Missing `libkrunfw`**: `libkrun` requires `libkrunfw.so.5` (a specialized minimal guest kernel payload built from `containers/libkrunfw`). Without `libkrunfw`, `libkrun` fails at runtime (`Can't load libkrunfw symbols`). Neither Dakota nor `freedesktop-sdk` packages `libkrunfw`, so it needs to be added as a dependency element.
- **`GPU=1` build flag requires `virglrenderer`**: `make BLK=1 NET=1 GPU=1 SND=1` enables virtio-gpu via rutabaga_gfx, which probes for `virglrenderer`. `freedesktop-sdk` does not provide `components/virglrenderer.bst`. Either package `virglrenderer` or drop `GPU=1` if GPU virtualization is not strictly required for the review runtime.
- **Offline Cargo sources missing**: Sandboxes lack network access. Compiling `libkrun` requires vendored crates declared as a `kind: cargo2` source block generated with `python3 files/scripts/generate_cargo_sources.py Cargo.lock`.
- **Install paths & missing headers**:
  - The Makefile outputs `target/release/libkrun.so.1.19.0`, so `install -Dm755 libkrun.so.1.19.0 ...` fails.
  - The current install commands do not install `include/libkrun.h` or `libkrun.pc`. Use `make PREFIX=%{prefix} DESTDIR="%{install-root}" install` so both the shared library and C headers are properly installed.

---

### 3. `crun-krun.bst` Build Failures
- **Missing header check**: `crun`'s `configure.ac` checks for `libkrun.h` via `AC_CHECK_HEADERS([libkrun.h], ..., AC_MSG_ERROR([*** Missing libkrun headers]))`. It will fail unless `libkrun.bst` installs headers into `%{includedir}`.
- **Missing `json-c` dependency**: `crun` 1.28 requires `json-c >= 0.14`. Add `freedesktop-sdk.bst:components/json-c.bst` to `build-depends` and `depends`.
- **Obsolete configure flag**: Drop `--enable-embedded-yajl` (no longer exists in crun 1.28).

---

### 4. `gocryptfs.bst` Non-Determinism & Sources
- **Offline Go modules**: Like `libkrun`, Go builds run with `GOPROXY=off` and need all upstream modules declared in `sources:`.
- **Timestamp / Git non-determinism**: Upstream `build.bash` executes `$(date +%Y-%m-%d)` and bakes it into `main.BuildDate`, and inspects `.git` for versioning. To comply with Dakota's deterministic build invariant (no timestamps/host metadata), set `SOURCE_DATE_EPOCH` or invoke `go build` directly with explicit `-ldflags`.

PR #1597, opened 2026-09-18
feat(packaging): add opensnitchd daemon

### LINKED ISSUE #1583: Package opensnitch and opensnitch-ui
### What would you like to see?

I'd like to see opensnitch daemon packaged and also its UI : https://github.com/evilsocket/opensnitch

The daemon should be disabled by default as it would not be useful for everyone.

### What does done look like?

User can either enable opensnitchd service or start it whenever they feel it useful.

With the UI, it will show information and ask deny/accept for the binary name / remote IP / remote PORT for some duration. This allow a fine grained real time monitoring of network connections, allowing to know what uses network and where, and ability to allow/deny it.

This is a really powerful fine grained firewall with no equivalent AFAIK, and it can't be installed without bundled it in the image.

### Affected area

GNOME / desktop

### Extra context

Requires libnetfilter-queue and libnfnetlink as dependencies for the daemon.

The UI requires PyQt6 to work.

I've been able to package opensnitchd but even with the help of copilot, I only came with a pretty horrible solution as opensnitchd is a Go program that does not ship a tarball with vendored dependencies.

On my local dakota, I have opensnitchd bundled in dakota image, and the UI working from its sources directory, so there are no breaker, but I struggle too much with bst at the moment to provide a good enough PR.

I'm not even sure opensnitch would be OK as a package in the base image. Maybe the UI could be packaged as a flatpak as a separate project? Other distros would be able to benefit from it, it only need to talk to an opensnitch instance over Unix socket or TCP (you can control daemons remotely)

rapenne-s:
Assisted-by: copilot with mai-code-1.1-flash model

This implements opensnitch daemon to fix https://github.com/projectbluefin/dakota/issues/1583

---

## Changes

This adds dependencies for the daemon of opensnitch github.com/evilsocket/opensnitch

The default configuration file is patched as this make it easier to use the flatpak version I made for the GUI in https://git.coopix.eu/public/flatpak-opensnitch-ui , otherwise the default unix socket is created by the GUI in /tmp/ which is forbidden by flatpak sandboxing, so I defaulted to using TCP to make the setup easier.

I am using it on my system, it is working fine, but although the daemon is packaged correctly, the extra flatpak program (that will never make it to flathub because they forbid this kind of GUI that depend on another service to be useful) feels clunky. I tried to package the UI in dakota, but I didn't succeed.

Honestly, I don't even know if there is an interest, in such situation I'd prefer to know it clearly, so I could stop spending hours on this :)

## Testing

- [x] `just validate` passes
- [x] `just boot-test` passes (automated boot smoke test)
- [x] `just lint` passes on a built image
- [x] `just boot-fast` or `just boot-vm` — desktop comes up, no regressions

## Checklist


ahmedadan:
@rapenne-s I am still thinking about adding this, I'll let you know when I think we can land it. 

Also, for future reference, if I attach myself to the issue, it means I'd like to take a closer look because we have other planned pieces in a similar area and I want to make sure we don't have a conflict.

ahmedadan:
@rapenne-s Looks like we need to add in some eBBF items for this to work properly. I'll have an actual review ready in a little bit here.

rapenne-s:
> @rapenne-s Looks like we need to add in some eBBF items for this to work properly. I'll have an actual review ready in a little bit here.

Indeed, eBPF support would increase opensnitch efficiency (as explained on [their wiki](https://github.com/evilsocket/opensnitch/wiki/monitor-method-ebpf)). But this requires adding their kernel module https://github.com/evilsocket/opensnitch/tree/master/ebpf_prog

The fallback [is using audit mode](https://github.com/evilsocket/opensnitch/wiki/monitor-method-auditd) as I understand.

elements/bluefin/opensnitchd.bst
```diff
@@ -0,0 +1,236 @@
...
+    export PATH="/tmp/protoc-bin:$PATH"
+
+    protoc -I../proto --go_out=../daemon/ui/protocol/ --go-grpc_out=../daemon/ui/protocol/ \
+      --go_opt=paths=source_relative --go-grpc_opt=paths=source_relative ../proto/ui.proto
+
+    go build -mod=vendor -o opensnitchd .
+
+  install-commands:
+  - |
+    install -Dm755 daemon/opensnitchd "%{install-root}%{bindir}/opensnitchd"
```
ahmedadan, line 215: This builds the daemon but not its eBPF programs. Please also build and install `opensnitch.o`, `opensnitch-procs.o`, and `opensnitch-dns.o` under `/usr/lib/opensnitchd/ebpf/`, using the same source revision. Dakota’s existing SDK LLVM can supply the build tools; no global toolchain pin is needed.

elements/bluefin/opensnitchd.bst
```diff
@@ -0,0 +1,236 @@
...
+  - |
+    sed -e 's|/usr/local/bin/opensnitchd|/usr/bin/opensnitchd|g' \
+        daemon/data/init/opensnitchd.service > opensnitchd.service.patched
+    install -Dm644 opensnitchd.service.patched \
+      "%{install-root}%{indep-libdir}/systemd/system/opensnitchd.service"
+  - |
+    install -d "%{install-root}%{sysconfdir}/opensnitchd/rules"
+    install -d "%{install-root}%{sysconfdir}/opensnitchd/tasks"
+    install -Dm644 daemon/data/default-config.json "%{install-root}%{sysconfdir}/opensnitchd/default-config.json"
```
ahmedadan, line 224: This configuration selects eBPF, but full process-event monitoring also requires `CONFIG_FTRACE_SYSCALLS`. It is disabled in both inspected Dakota kernel configurations, so we need a coordinated change in `files/linux/dakota-config.sh` and runtime verification before treating this mode as supported.

elements/bluefin/opensnitchd.bst
```diff
@@ -0,0 +1,236 @@
...
+    sed -i 's|"Address":"unix:///tmp/osui.sock"|"Address":"127.0.0.1:50051"|g' \
+      "%{install-root}%{sysconfdir}/opensnitchd/default-config.json"
```
ahmedadan, line 226: Could we use a private runtime Unix socket instead? The Flatpak manifest already exposes `xdg-run/opensnitch`, and I verified that this transport works without network sharing. That avoids trusting whichever local process binds this unauthenticated TCP port; the GUI launcher and explicit user setup would need to configure matching endpoints.

elements/bluefin/opensnitchd.bst
```diff
@@ -0,0 +1,236 @@
...
+
+sources:
+# BuildStream sandboxes are network-isolated. Each go_module source stages its
+# dependency under vendor/<module> and appends to the generated
+# vendor/modules.txt. Keep that output intact and only expose it to the daemon
+# checkout via a local vendor symlink.
+- kind: git_repo
+  url: github:evilsocket/opensnitch.git
+  track: master
+  ref: a1353848ba1b660320e90cefea782c3fba272c00
```
ahmedadan, line 15: This revision identifies as OpenSnitch 1.9.0, while the linked Flatpak manifest pins UI 1.8.0. Please coordinate a matched daemon/UI version and test it together, rather than make cross-version compatibility part of the supported setup.

elements/bluefin/opensnitchd.bst
```diff
@@ -0,0 +1,236 @@
...
+    go build -mod=vendor -o opensnitchd .
+
+  install-commands:
+  - |
+    install -Dm755 daemon/opensnitchd "%{install-root}%{bindir}/opensnitchd"
+  - |
+    sed -e 's|/usr/local/bin/opensnitchd|/usr/bin/opensnitchd|g' \
+        daemon/data/init/opensnitchd.service > opensnitchd.service.patched
+    install -Dm644 opensnitchd.service.patched \
+      "%{install-root}%{indep-libdir}/systemd/system/opensnitchd.service"
```
ahmedadan, line 220: For Dakota’s proposed session-scoped behavior, this service also needs logout lifecycle integration. Upstream has graceful cleanup, but this unit does not connect shutdown to the controlling desktop session. Please verify that logout removes OpenSnitch’s active rules and probes without touching other firewall rules, and test crash recovery separately.

ahmedadan (commented):
Thanks @rapenne-s for working through the packaging. I built the unchanged daemon package on x86_64. Please address the inline packaging and integration points and add runtime coverage for the supported architectures.

Danathar:
Thanks for packaging this, and for the clear write-up of the Flatpak trade-offs.

I went through the element while ahmedadan works through the eBPF and socket questions. Things that look right to me: every source is pinned to a commit, the preset ships `disable opensnitchd.service` so nothing changes for users who do not opt in, and the two `sed` edits match the upstream files at commit a1353848 exactly.

One small thing on top of ahmedadan's points: upstream is GPL-3.0, and other elements here (`syncthing.bst`, `umotd.bst`) install the LICENSE file under `/usr/share/licenses/<name>/`. It would be good to do the same for opensnitchd.

Whether this lands is ahmedadan's call given the related work he mentioned; I just wanted you to know it is being looked at.

rapenne-s:
Thank you both for your reviews, I'll sort this out, it may take me some time.

elements/bluefin/opensnitchd.bst
rapenne-s: Would it be ok to listen on a default path for uid 1000 (it is likely that dakota users running opensnitch use a single user), this would allow to choose `unix:///run/user/1000/opensnitch//osui.sock` in the UI which is one of the default path. It should work with flatpak too.

Issue #1409, opened 2026-08-26
gaming images: publish SPDX SBOM referrers


castrojo:
## What

The published gaming images do not attach `application/vnd.spdx+json` SBOM referrers, so downstream consumers cannot source the shipped OGC kernel version without falling back to source-tree guesses.

## Reproduction

Verified 2026-08-26:

- `ghcr.io/projectbluefin/dakota-gaming:testing` → `sha256:aff99512138bb5cb6e58926373a1d1bb794697914808845b4cb7f22f1cd6e585`
- `ghcr.io/projectbluefin/dakota-nvidia-gaming:testing` → `sha256:a80845e95e0b2fa32941f8662ba68b7798e5142735a9ed6efe464ab72031c902`

```console
oras discover --artifact-type application/vnd.spdx+json --format json ghcr.io/projectbluefin/dakota-gaming:testing
oras discover --artifact-type application/vnd.spdx+json --format json ghcr.io/projectbluefin/dakota-nvidia-gaming:testing
```

Both return no SPDX referrer. An unfiltered `oras discover --format json` shows only Sigstore/SLSA provenance referrers.

For comparison, `dakota:latest` and `dakota-nvidia:latest` do publish SPDX SBOM referrers.

## Fix

Attach the generated SPDX SBOM to both gaming image manifests using the same artifact type and publication path as the non-gaming variants. Ensure the SBOM preserves the `core/linux-ogc.bst` element/source version so downstream automation can identify the shipped OGC kernel.

Automatable: yes

Danathar:
Before this can happen for the gaming images: SBOMs aren't being published for the base images either. `publish-sbom` has failed on every Publish since 2026-06-16 (#1608, fix in #1609). Once that's green again, the gaming variants are the incremental step this issue describes.

PR #1609, opened 2026-09-19
fix(sbom): give the buildstream-sbom container /dev/fuse so fetch can stage

### LINKED ISSUE #1608: ci: publish-sbom has failed on every run since June — buildstream-sbom container has no /dev/fuse
## What happens

`publish-sbom` in `Publish Bluefin dakota` has failed on every run since **2026-06-16** — three months — for both the `default` and `nvidia` variants. Because the matrix sets `continue-on-error: true`, Publish stays green and no SBOM referrer has been attached to any published image in that time.

Every failure has the same signature, at the very first `fetch`:

```
[fetch:freedesktop-sdk.bst] BUG  Fetch
    ...
    File ".../buildstream/_cas/cascache.py", line 511, in stage_directory
      response = next(response_stream)
    grpc._channel._MultiThreadedRendezvous: <_MultiThreadedRendezvous of RPC that terminated with:
      status = StatusCode.INTERNAL
      details = "std::runtime_error exception thrown at [buildboxcommon_fusestager.cpp:149],
                 errMsg = "The FUSE stager child process unexpectedly died with exit code 2""
error: recipe `sbom` failed with exit code 1
```

Sampled runs: 35434468946 and 35428241200 (2026-09-19), 35349468461 (09-18), runs on 08-02, 06-17 and 06-16 — all identical. Runs before 06-16 show the job as `skipped`, so I could not find a successful one.

## Cause

#745 (`a74b602`, 2026-06-08) removed `--privileged` from the `sbom` recipe's `podman run` in the `Justfile`, on the reasoning that `bst show` and `buildstream-sbom` are read-only and need no FUSE mounts. The log shows `buildstream-sbom` does trigger a source fetch, and BuildStream stages fetched sources through `buildbox-fuse` — which needs `/dev/fuse` and the capabilities to mount it. The first failing run is one week after that commit.

The `bst` wrapper recipe a few lines above still runs with `--privileged --device /dev/fuse` and works on the same `ubuntu-24.04` runners every day, so the flags themselves aren't a problem on the current runner image (the crun probe issue #745 hit was on a resolute/26.04 experiment; #749 then switched the sbom container to `runc`).

## Fix

Give the `buildstream-sbom` container the same `--privileged --device /dev/fuse` the `bst` wrapper uses. PR to follow. The job is non-blocking, so a wrong fix cannot break Publish — but it also means the only real verification is the next Publish on `testing`.

## Related

- #1409 asks for SBOM referrers on the gaming images; that is moot until the base variants publish one again.

Danathar:

#1608: `publish-sbom` has failed on every Publish run since 2026-06-16, so no SBOM referrer has been attached to a published image in three months. It's `continue-on-error`, which is why Publish stayed green.

Every run dies at the first fetch (`freedesktop-sdk.bst`) with

```
errMsg = "The FUSE stager child process unexpectedly died with exit code 2"
```

## Cause and fix

@castrojo — this reverses part of your #745, so flagging it plainly. #745 dropped `--privileged` from the `sbom` recipe's container because `buildstream-sbom` looked read-only. The log shows it does fetch sources, and BuildStream stages fetched sources through `buildbox-fuse`, which needs `/dev/fuse` and the capabilities to mount it. The first failing run is one week after #745.

This gives the container the same `--privileged --device /dev/fuse` the `bst` recipe a few lines above runs with — and that works on the same `ubuntu-24.04` runners every day. The `--runtime runc` from #749 is kept.

## Testing

- `just --evaluate` parses the Justfile.
- The job can't be exercised on a PR (it only runs from Publish on `testing`/`next`), so the real verification is the first Publish after this lands: `publish-sbom (default)` and `(nvidia)` should go green and `oras discover` on the new digest should show an `application/spdx+json` referrer. If it still fails with a different error, that's new information rather than this fix being wrong.

Closes #1608
Refs #1409

https://claude.ai/code/session_01VXwvRu29c2MkwftzkgAHcC


ahmedadan:
@Danathar This is going in as part of a larger PR with other CI changes, so I'm going to close this as duplicate.

PR #1612, opened 2026-09-19
chore(kernel): bump next to 7.3-rc3 and OGC 7.2.6-ogc1

ahmedadan:
## Summary

The next stream's kernels had drifted on both sides: the normal kernel sat on v7.3-rc1 while upstream is at rc3, and the gaming kernel was still on v7.2.1-ogc4, a full five point releases behind what testing runs. Both elements are in the sync-next `NEXT_OWNED` overlay, so this has to land on next directly rather than flow through from testing. It also puts `-Werror` back on the gaming kernel, matching the testing PR.

1. **`elements/core/linux-fdsdk.bst` moves from v7.3-rc1 to v7.3-rc3.** The two carried patches in `patches/linux` (the gcc15 fix and the riscv fu740 compatible string) were dry-run applied against the rc3 files they touch and both go on clean.
2. **`elements/core/linux-ogc.bst` moves from v7.2.1-ogc4 to v7.2.6-ogc1 and `ogc-localversion` becomes `-ogc1`.** This is the same OGC tag testing is moving to in its own PR, so both streams end up on the same gaming kernel. OGC has no 7.3 branch or tag yet, and a dry-run rebase of their patchset onto rc3 showed the AMD HDMI 2.1 VRR stack conflicting with 7.3's new FreeSync module, so carrying our own 7.3 OGC kernel is not worth it until they ship one.
3. **`CONFIG_WERROR` is enabled again through the expectation gate.** #1473 dropped it because the OGC tree was not warning-clean, but the local build logs show exactly one warning in the whole tree, an unused mutex in the out-of-tree aw87xxx speaker amp driver. This goes back to `enable WERROR` so a new warning in a future tag bump fails the build instead of shipping.
4. **New patch `0002-ASoC-aw87xxx-only-define-g_dsp_lock-on-DSP-platforms.patch` fixes that one warning.** The mutex is only taken inside the MTK and Qualcomm DSP blocks that x86 never compiles, so its definition is guarded with the same condition. Verified to apply clean against v7.2.6-ogc1.
5. **The carried ayn-ec patch stays.** The macro collision from [OpenGamingCollective/linux#98](https://github.com/OpenGamingCollective/linux/issues/98) is still present in 7.2.6-ogc1 and would fail the build again under `-Werror` without it.

**Verified:** both tags resolve to the pinned commits (`fd73f4a6` for rc3, `d7173986` for ogc1), the `patches/linux` queue applies to rc3, and both `patches/linux-ogc` patches apply to ogc1. Not built and not booted, nvidia-drivers has not been compiled against rc3, and this is the first `-Werror` build of the ogc1 tree, the next build will exercise all three.

Related: #1473, OpenGamingCollective/linux#98


castrojo (approved):

Danathar (approved):
What I checked: `fd73f4a66598…` is the commit behind torvalds' annotated `v7.3-rc3` tag, and `d71739861de0…` is OGC's `v7.2.6-ogc1`; both refs match. The `patches/linux` queue and both `patches/linux-ogc` patches were checked by hand against the new trees per the description.

Same caveat as #1613, and it applies twice here: neither kernel has been built (rc3 and the first `-Werror` ogc1 compile), nvidia-drivers hasn't been compiled against rc3, and there is no PR build path in this repo — `pr-image` is unreachable on pull requests because it depends on a `build` job that only runs on push. `next` is the rolling stream, so the first push build is the proof and a red one is the expected way to find out. castrojo already approved.

PR #1613, opened 2026-09-19
chore(kernel): bump OGC gaming kernel to 7.2.6-ogc1 on testing

ahmedadan:
## Summary

The gaming variants on testing run v7.2.4-ogc3 while the normal kernel is already on 7.2.6, so gaming images were two stable point releases behind the rest of the stream. OGC tagged [v7.2.6-ogc1](https://github.com/OpenGamingCollective/linux/releases/tag/v7.2.6-ogc1) on 2026-09-19 and this moves testing onto it. It also puts `-Werror` back on the gaming kernel, which #1473 had switched off.

1. **`elements/core/linux-ogc.bst` pins v7.2.6-ogc1 and sets `ogc-localversion` to `-ogc1`.** This is a stable point bump within the 7.2 series, not a series jump, so the nvidia-drivers build-tree rule in the element does not apply. `uname -r` becomes `7.2.6-ogc1`.
2. **The OGC patchset delta over ogc3 is small.** On top of stable 7.2.4 to 7.2.6 it adds panel orientation quirks for the Zeenix Lite and Pro, a FreeSync minimum range fix for quirky panels, and RGB LED naming tweaks in hid-oxp and hid-msi.
3. **`CONFIG_WERROR` is enabled again through the expectation gate.** #1473 dropped it because the OGC tree was not warning-clean, but the local build logs show exactly one warning in the whole tree, an unused mutex in the out-of-tree aw87xxx speaker amp driver. That is not enough reason to let new warnings from the gaming patchset land silently, so this goes back to `enable WERROR` and the configure stage now fails if anything drops it.
4. **New patch `0002-ASoC-aw87xxx-only-define-g_dsp_lock-on-DSP-platforms.patch` fixes that one warning.** The mutex is only taken inside the MTK and Qualcomm DSP blocks that x86 never compiles, so its definition is guarded with the same condition. Verified to apply clean against v7.2.6-ogc1.
5. **The carried ayn-ec patch stays.** The local `DEVICE_ATTR_*_NAMED` macros are unchanged in the 7.2.6-ogc1 tree, so the collision reported in [OpenGamingCollective/linux#98](https://github.com/OpenGamingCollective/linux/issues/98) is still there and would now fail the build again without it.

**Verified:** the tag resolves to `d7173986` and the ref string matches the tracker format, both carried patches were checked against the new tree by hand, and the single-warning count comes from the v7.2.4-ogc3 build logs of the current testing artifact. Not built and not booted, the PR build is the first compile of this tree with `-Werror` on and will be the proof that ogc1 introduced nothing new.

Related: #1473, OpenGamingCollective/linux#98


castrojo (approved):

Danathar (approved):
What I checked: `d71739861de0…` is OpenGamingCollective/linux's `v7.2.6-ogc1` tag (lightweight, resolved directly), and the ref string matches the tracker format. The carried patch is scoped to the one `-Werror` failure identified from the v7.2.4-ogc3 build log (unused `g_dsp_lock` on non-DSP platforms).

What is *not* checked, and can't be here: the kernel hasn't been built with `-Werror` on against this tree — the description says so — and dakota has no PR build path (`pr-image` requires `needs.build.result == 'success'`, but `build` only runs on push/schedule/dispatch, so it is skipped on every PR regardless of the `build` label). The push to `testing` after merge is the first compile. If ogc1 carries a second warning, that build goes red and this needs a follow-up patch; `stable` is unaffected either way because publish gates on the build. castrojo already approved.

PR #1622, opened 2026-09-20
feat(countme): remove local countme client and units

### LINKED ISSUE #1524: fix(countme): point countme reporting to countme.projectbluefin.io/metalink
## Summary

`files/countme/dakota-countme` in this repository currently targets `https://mirrors.fedoraproject.org/metalink?repo=fedora-${FEDORA_VER}&arch=${ARCH}&countme=${COUNTME}`.

Because this sends pings to Fedora upstream mirrors using a Fedora base repo format, Dakota's adoption remains unmeasurable on Project Bluefin's own dashboard (`docs.projectbluefin.io/analytics`).

## Required Implementation

1. Update the countme URL in `files/countme/dakota-countme` to target `https://countme.projectbluefin.io/metalink?repo=dakota&tag=${IMAGE_TAG}&flavor=${IMAGE_FLAVOR}&arch=${ARCH}&countme=${COUNTME}`.
2. Honor `/etc/projectbluefin/countme/disabled` and `/etc/dakota-countme/disabled` for opt-outs.
3. Keep user agent as `dakota-countme (${IMAGE_NAME} ${VERSION_ID}; ${IMAGE_FLAVOR}; Linux.${ARCH})` with no machine ID or persistent identifier.

Ref: `projectbluefin/documentation` ADR 0006.

castrojo:

Remove Dakota's legacy local countme reporter and unit files. We'll put this in common, and Dakota will just ingest common normally.

---

## Changes

- Remove `elements/bluefin/countme.bst`
- Remove `files/countme/` (`dakota-countme`, `bluefin-countme.service`, `bluefin-countme.timer`)
- Drop `bluefin/countme.bst` from `elements/bluefin/deps.bst`
- Clean up runtime countme references in `include/os-release.yml` and workstation skill docs

## Testing

- [x] `just validate` passes

## Checklist

- [x] New BST elements wired into `deps.bst` (removed obsolete dependency)

Closes #1524


Danathar (commented):
This is the removal half of #1561, and it targets `testing`, which fixes the branch problem from there. It cannot land on its own yet, for one reason.

At this head `elements/bluefin/common.bst` still pins `v2026.08-112-g8df20e8`. Neither that ref nor common `main` contains `bluefin-countme.service`, `bluefin-countme.timer` or `/usr/libexec/bluefin-countme`; they only exist on common#1112, which is still open with changes requested. So if this merged today, Dakota would stop pinging Fedora and nothing would ping countme.projectbluefin.io either. No counting at all until common#1112 lands and someone bumps the common ref here.

I think the order is: common#1112 merges, then this PR bumps `common.bst` (`bst source track bluefin/common.bst`) in the same change that removes the local files. Then the removal is safe and I will approve. The removal itself looks correct: `grep -rni countme` on the PR tree comes back empty, so nothing is left pointing at the deleted script.

Two smaller things:

- common#1112 enables the timer only through a preset (`03-bluefin-countme.preset`). As far as I can tell presets apply on first boot only, so existing Dakota installs that upgrade to an image without the old static `timers.target.wants/bluefin-countme.timer` link would never have the shared timer enabled. Either keep a static link in Dakota pointing at common's timer (what #1561 did), or have common ship the link. Worth settling before the ref bump.
- "Closes #1524" will auto-close an issue that asks Dakota to report to countme.projectbluefin.io, while this PR by itself removes reporting. "Refs #1524" until the common bump lands. The body should also say plainly that the Fedora "Dakota" ping goes away.

CI note: the one cancelled `Validate` run is a concurrency cancel; the second run for the same head passed, and the image build is queued behind the global build group.

